Issue #71, opened 2022-04-14
Validator set from the Sparrow's perspective


Vizualni:
# Depends on
 - #66

# Soft dependency on

- #48  (could be "faked" without this, but would be configurable for a closed alpha version test)
  - #81
 
# Background

This ticket should be done after #66. Now chain supports registering conductors. From the conductor's perspective we should register as soon as it starts and should add external chain id info that the runner knows about. Also, when there is a config change (when conductor gets restarted) it should check the state and apply it to the chain (k8s style)

# Done when

- [x] conductor can register itself with the chain
- [x] conductor can send external chain info to our chain

PR #98, opened 2022-05-02
Fix when trying to use validator's key to sign messages

Vizualni:
# Related Github tickets

- [a list of github issues that are relevant for this PR](https://github.com/palomachain/paloma/issues/71)

# Background

Initial assumption (due to the fact that I am still learning cosmos ecosystem) was that we can use validator's keys for signing messages. We can't :) (I mean, we can, but that would be a hack^2). 
# Testing completed

- [x] wrote tests
- [x] tested this with Sparrow to see that it works as intended


taariq:
What does Gravity use for validators to sign messages?

Vizualni:
This is something that I've taken from gravity
https://github.com/VolumeFi/gravity-bridge/blob/0bfefb9238213d92f6d950729e1824cb7e124ce0/module/x/gravity/types/msgs.pb.go#L695-L697. They register the eth accounts which they use to verify signatures. I've taken that approach and applied it here.

PR #134, opened 2022-05-19
Add quoting in setup-chain-validator, set -eu -o pipefail

measure-fi:
# Related Github tickets

- None

# Background

`set -euo pipefail` adds extra error handling in a bash script, quoting prevents some simple errors.
We should prefer to have these in almost every bash script.

# Testing completed

- [x] Manual testing


scripts/setup-chain-validator.sh
```diff
@@ -1,19 +1,13 @@
-#!/usr/bin/env bash
-
-set -ex
+#!/bin/bash
```
Vizualni, line 1: Please revert the shebang. You can't know where person has their bash shell and maybe they don't want you to use the /bin/bash one but rather one that they've defined.

Vizualni (changes requested):
lgtm, but there is one request to change

Vizualni (approved):

PR #152, opened 2022-05-23
Add Frontend request to issue templates

verabehr:
Adding a template that the whole org can use when FE and/or Design work is needed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/frontend-request.md
+++ b/.github/ISSUE_TEMPLATE/frontend-request.md
@@ -1,0 +1,20 @@
+---
+name: Frontend request
+about: Request for Frontend and Design work
+title: ''
+labels: Ui/UX
+assignees: joycehermosilla, verabehr
+
+---
+
+**Describe what you need**
+A clear and concise description of what you need.
+
+**What is this needed for**
+Explain why this is needed.
+
+**Desired launch date**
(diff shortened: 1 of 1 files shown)


rootedbox (approved):

.github/ISSUE_TEMPLATE/frontend-request.md
```diff
@@ -0,0 +1,20 @@
...
+name: Frontend request
+about: Request for Frontend and Design work
+title: ''
+labels: Ui/UX
+assignees: joycehermosilla, verabehr
+
+---
+
+**Describe what you need**
+A clear and concise description of what you need.
```
measure-fi, line 11: [Currently these two lines get rendered as one line](https://github.com/palomachain/paloma/blob/ffd3078339c78f0abcfb4e76d608060cae24c1e8/.github/ISSUE_TEMPLATE/frontend-request.md), I'm not sure that's intended.

measure-fi: It's also not clear from context here what text is intended to be verbatim and what text is meant to be replaced. I think this section e.g. should be:

```
## Description

_A clear and concise description of what you need._
```

which renders as:

## Description

_A clear and concise description of what you need._

Issue #143, opened 2022-05-20
Tracking issue for private testnet 20/05/2022


Vizualni:
# Background

Our first private testnet. Please when creating new bug reports, make sure to link this issue in the bug report so that we have this all in one place.

# Bug tickets

- _fill in later_

measure-fi:
@Vizualni Added some tasks to the list. Feel free to edit them if they are wrong in some particular.

measure-fi:
@Vizualni Is moniker necessary to set or can we elide it? What does moniker actually affect?

Vizualni:
@measure-fi please create a ticket for each one of those issues that you've mentioned

measure-fi:
@Vizualni @verabehr Is this Issue superseded by the [testnet label](https://github.com/orgs/palomachain/projects/1/views/4?filterQuery=label%3ATestnet+is%3Aopen)? Can we close?

verabehr:
I'm fine closing.

PR #195, opened 2022-06-06
Make a new (uncorrupted) testnet based on an up-to-date palomad

measure-fi:
https://github.com/palomachain/paloma/issues/143

# Background

Attempting to upgrade the palomad version resulted in corruption. Making a new testnet so people have something to connect to at least.

# Testing completed

- [x] previous testnet connection only


Vizualni (approved):
lgtm

PR #199, opened 2022-06-07
Update testnet instructions with addrbook

measure-fi:
https://github.com/palomachain/paloma/issues/143

# Background

Continue to evolve testnet creation instructions.

# Testing completed

- [ ] Manual testnet setup.

[Rendered.](https://github.com/palomachain/paloma/blob/chase-new-testnet/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,10 +61,11 @@ MONIKER="$(hostname)"
 palomad init "$MONIKER"
 ```
 
-Copy the `genesis.json` file of the testnet we wish to connect to
+Copy the configs of the testnet we wish to connect to
 
 ```shell
 wget -O .paloma/config/genesis.json https://raw.githubusercontent.com/palomachain/testnet/master/livia/genesis.json
+wget -O .paloma/config/addrbook.json https://raw.githubusercontent.com/palomachain/testnet/master/livia/addrbook.json
 ```
 
 Next you will need to add a new set of keys to the new machine.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,13 +128,17 @@ ConditionPathExists=/usr/local/bin/palomad
 [Service]
 Type=simple
 Restart=always
+RestartSec=5
 WorkingDirectory=~
 ExecStartPre=
-ExecStart=/usr/local/bin/palomad start --p2p.persistent_peers da10f3078f78dabbb38b59df2c90230716f3149c@157.245.76.119:26656
+ExecStart=/usr/local/bin/palomad start --p2p.persistent_peers f64dd167410a242c993648faa6406edf74a7f4b7@157.245.76.119:26656
```
measure-fi, line 134: @Vizualni Is there a way to remove the `--p2p...` part? I thought adding addrbook.json would do this.

measure-fi:
And stop is not a `palomad` command as far as I can tell. It can just be killed at any time. `stop` is a systemd command, but that's not something we need to worry about.

verabehr (approved):
works and easier than last time. good to go.

verabehr:
> > we should update the description for getting funds. Currently we need to ask for it
> 
> * [Rewrite "query balance" in testnet instructions to allow being run from anywhere #203](https://github.com/palomachain/paloma/issues/203)
> 
> Unless you mean we need an issue for a "faucet" api, in which case yes, definitely.

No, I mean, if we have others (who are not in our zoom) come in and join the testnet. Wouldn't they have to reach out to someone to get the funds? I don't think they can run the command to send doves themselves?

measure-fi:
@verabehr Yes, the situation of messaging someone for grain is unacceptable. Moved to https://github.com/palomachain/paloma/issues/205

README.md
Vizualni: It should work with the address book. I haven't really played with the address book so I wouldn't know at this point :(. Initially I wanted to use addresbook, but it didn't work for me so I've decided to use the --p2p arguments instead and leave the addressbook for future Matija to figure out :)

Issue #203, opened 2022-06-07
Rewrite "query balance" in testnet instructions to allow being run from anywhere


measure-fi:
# Background

Currently it's directed to run from the main node `palomad query bank balances`. This could instead read:

```
palomad query bank balances --node=tcp://157.245.76.119:26657 $VAL
```

and it would work from anywhere.

# Done when

- [ ] Docs are updated

PR #204, opened 2022-06-07
Allow balances query in testnet setup to be run from anywhere

measure-fi:
Fixes https://github.com/palomachain/paloma/issues/203

# Background

Previously this query could only be run from a node already in the network. Now it should be runnable from anywhere.

# Testing completed

- [x] The new query runs on nodes not running the network.


verabehr (approved):
tested - works for me

PR #237, opened 2022-06-17
Initialize new testnets with updated values

measure-fi:
[Public Testnet EPIC](https://www.notion.so/volumefi/EPIC-06-17-Paloma-Public-Testnest-1-d11ce23143c7475a9157acd224933d81)

# Background

We want a number of different values for testnet initialization, add them here.

# Testing completed

- [x] Test initializing a new chain

### DIFF
--- a/scripts/setup-volume-testnet.sh
+++ b/scripts/setup-volume-testnet.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+set -euo pipefail
+set -x
+
+if ! which jq > /dev/null; then
+  echo 'command jq not found, please install jq'
+  exit 1
+fi
+
+CHAIN_ID="paloma"
+VALIDATOR_STAKE_AMOUNT=100000000grain
+
+palomad init my_validator --chain-id "$CHAIN_ID"
+
+pushd ~/.paloma/config/
+sed -i 's/^keyring-backend = ".*"/keyring-backend = "test"/' client.toml
+sed -i 's/^minimum-gas-prices = ".*"/minimum-gas-prices = "0.001grain"/' app.toml
+sed -i 's/^laddr = ".*:26657"/laddr = "tcp:\/\/0.0.0.0:26657"/' config.toml
+jq ".chain_id = \"${CHAIN_ID}\"" genesis.json > genesis.json.tmp && mv genesis.json.tmp genesis.json
+jq 'walk(if type == "string" and .. == "stake" then "grain" else . end)' genesis.json > genesis.json.tmp && mv genesis.json.tmp genesis.json
+popd
+
+name="chase"
+echo "$MNEMONIC" | palomad keys add "$name" --recover
+address="$(palomad keys show "$name" -a)"
+
+palomad add-genesis-account "$address" 250000000grain
+palomad gentx "$name" "$VALIDATOR_STAKE_AMOUNT" --chain-id "$CHAIN_ID"
+
+init() {
+  name="$1"
+  address="$2"
+  amount="${3:-250000000grain}"
+
+  palomad add-genesis-account "$address" "$amount"
+}
+
+init faucet paloma167rf0jmkkkqp9d4awa8rxw908muavqgghtw6tn 500000000grain
+init jason paloma1mre80u0mmsdpf3l2shre9g4sh7kp9lxu5gtlql
+init taariq paloma1k4hfe8cqdzy6j0t7sppujhrplhp8k5tglf8v46
+init vera paloma1nvhryyg63e8vw9t5sx9ht8l77jzgjvrq6jcdnp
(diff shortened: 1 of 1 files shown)


scripts/setup-volume-testnet.sh
```diff
@@ -0,0 +1,36 @@
...
+  name="$1"
+  address="$2"
+  
+  palomad add-genesis-account "$validator" 250000000grain
+  palomad gentx "$name" "$VALIDATOR_STAKE_AMOUNT" --chain-id "$CHAIN_ID"
+}
+
+init_validator chase paloma1x6m0d47lcdka7mrg7997u80cyq60yg5adeekvx
+init_validator jason paloma1x6m0d47lcdka7mrg7997u80cyq60yg5adeekvx
+init_validator taariq paloma1nvhryyg63e8vw9t5sx9ht8l77jzgjvrq6jcdnp
```
measure-fi, line 33: @taariq What address should be used here (currently just giving Vera double)

scripts/setup-volume-testnet.sh
```diff
@@ -0,0 +1,36 @@
...
+init_validator() {
+  name="$1"
+  address="$2"
+  
+  palomad add-genesis-account "$validator" 250000000grain
+  palomad gentx "$name" "$VALIDATOR_STAKE_AMOUNT" --chain-id "$CHAIN_ID"
+}
+
+init_validator chase paloma1x6m0d47lcdka7mrg7997u80cyq60yg5adeekvx
+init_validator jason paloma1x6m0d47lcdka7mrg7997u80cyq60yg5adeekvx
```
measure-fi, line 32: @rootedbox What address should we use here (currently just giving me double)?

scripts/setup-volume-testnet.sh
```diff
@@ -0,0 +1,39 @@
...
+init_validator() {
+  name="$1"
+  address="$2"
+  
+  palomad add-genesis-account "$address" 250000000grain
+  palomad gentx "$name" "$VALIDATOR_STAKE_AMOUNT" --chain-id "$CHAIN_ID"
+}
+
+# Faucet
+palomad add-genesis-account paloma1fgy8ghc70swdzzpa4xpk544xs7q8xh3zfkzal2 4000000000grain
```
measure-fi, line 32: Do we want less in the faucet and any excess (up to 5B) stored in a 6th account?

scripts/setup-volume-testnet.sh
```diff
@@ -0,0 +1,39 @@
...
+jq 'walk(if type == "string" and .. == "stake" then "grain" else . end)' genesis.json > genesis.json.tmp && mv genesis.json.tmp genesis.json
+popd
+
+VALIDATOR_STAKE_AMOUNT=100000000grain
+init_validator() {
+  name="$1"
+  address="$2"
+  
+  palomad add-genesis-account "$address" 250000000grain
+  palomad gentx "$name" "$VALIDATOR_STAKE_AMOUNT" --chain-id "$CHAIN_ID"
```
measure-fi, line 28: @Vizualni Currently getting `Error: failed to fetch 'chase' from the keyring: chase.info: key not found` from this line, which makes sense. How do I add external keys without importing the private key?

measure-fi: This makes sense. I think the "chase" validator will have to execute this line and I'll make a new one every script invocation.

scripts/setup-volume-testnet.sh
```diff
@@ -0,0 +1,43 @@
...
+sed -i 's/^laddr = ".*:26657"/laddr = "tcp:\/\/0.0.0.0:26657"/' config.toml
+jq ".chain_id = \"${CHAIN_ID}\"" genesis.json > genesis.json.tmp && mv genesis.json.tmp genesis.json
+jq 'walk(if type == "string" and .. == "stake" then "grain" else . end)' genesis.json > genesis.json.tmp && mv genesis.json.tmp genesis.json
+popd
+
+name="chase"
+echo "$MNEMONIC" | palomad keys add "$name" --recover
+address="$(palomad keys show "$name" -a)"
+
+palomad add-genesis-account "$address" 4000000000grain
```
measure-fi, line 27: @verabehr As written this account has 4Ggr, the below accounts get 800Mgr total. What would you like the distribution to be?

verabehr: ok, so let's change the faucet to give out 100 000grain per request. 
us for (jason, taariq, you, me) each get 250000000 as you have it. 
the faucet starts out with 500 000 000grain as you have it
that gives us 1 500 000 000grain in total.

measure-fi: Done.

Issue #247, opened 2022-06-20
Update paloma-testnet-5 with more new values


measure-fi:
# Background

Genesis file still needs some edits.

# Done when

- https://github.com/palomachain/paloma/pull/249
- [x] Updated Validator Minimum for staking: 1 GRAIN
- [x] Updated Denoms for GRAINs: 1 GRAIN == 1,000,000 ugrain
  - Confirmed in https://github.com/palomachain/paloma/issues/224

PR #249, opened 2022-06-20
Choose a new chain-id when we make a new chain.

measure-fi:
https://github.com/palomachain/paloma/issues/247

# Background

We cannot keep reusing the same chain-id.

# Testing completed

- [x] manual review

### DIFF
--- a/scripts/setup-volume-testnet.sh
+++ b/scripts/setup-volume-testnet.sh
@@ -7,8 +7,19 @@ if ! which jq > /dev/null; then
   exit 1
 fi
 
-CHAIN_ID="paloma"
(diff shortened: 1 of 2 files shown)


scripts/setup-volume-testnet.sh
```diff
@@ -7,7 +7,10 @@ if ! which jq > /dev/null; then
   exit 1
 fi
 
-CHAIN_ID="paloma"
+if [[ -z "${CHAIN_ID:-}" ]]; then
+  echo 'CHAIN_ID required'
```
Vizualni, line 11: exit 1?

PR #250, opened 2022-06-20
Make ugrain default denom

Vizualni:
# Related Github tickets

# Background

# Testing completed


measure-fi:
@Vizualni Is https://github.com/palomachain/paloma/blob/master/scripts/setup-volume-testnet.sh#L20 still necessary as of this PR?

PR #266, opened 2022-06-24
Exit readme to reflect current faucet and staking realities

measure-fi:
https://github.com/palomachain/paloma/issues/247

# Background

There is no longer a command line accessible faucet. Staking command also needed to be tweaked, and mentioned as optional.

# Testing completed

- [x] manual review


verabehr (approved):
lgtm

PR #269, opened 2022-06-24
Set max_validators to 125 by default

measure-fi:
https://github.com/palomachain/paloma/pull/268

# Background

Normal default is 100, we would like more people to be able to be validators on the testnet.

# Testing completed

- [x] manual review


Vizualni (commented):
We will add this a bit differently. Not through this, but through default genesis

measure-fi:
@Vizualni Do you want to do the same thing for this line

```
jq-i 'walk(if type == "string" and .. == "stake" then "grain" else . end)' genesis.json
```

(which probably needs to use `000000ugrain` anyway)

Issue #267, opened 2022-06-24
Configuration updates for paloma-testnet-6


measure-fi:
# Background

Tracking small edits needed.

# Done when

- https://github.com/palomachain/paloma/pull/276
- https://github.com/palomachain/paloma/pull/278
- https://github.com/palomachain/paloma/pull/279

PR #279, opened 2022-06-28
Change seed values for initial validators for testnet

measure-fi:
https://github.com/palomachain/paloma/issues/267

# Background

We want these values to be different.

# Testing completed

- [x] Manual review

### DIFF
--- a/scripts/setup-volume-testnet.sh
+++ b/scripts/setup-volume-testnet.sh
@@ -17,10 +17,6 @@ if [[ -z "${MNEMONIC:-}" ]]; then
   exit 1
 fi
 
-VALIDATOR_STAKE_AMOUNT=100000000000000ugrain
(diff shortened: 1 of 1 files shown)


scripts/setup-volume-testnet.sh
```diff
@@ -37,12 +33,19 @@ sed -i 's/^laddr = ".*:26657"/laddr = "tcp:\/\/0.0.0.0:26657"/' config.toml
 jq-i ".chain_id = \"${CHAIN_ID}\"" genesis.json
 popd
 
+GR=000000ugrain
+MGR="000000${GR}"
+
+INIT_AMOUNT="5${MGR}"
```
taariq, line 39: @measure-fi Let's make sure that this address only validates 1MGR.

taariq (approved):
Super excited to see how this testnet works! LFG!

Issue #300, opened 2022-07-05
Configuration updates for paloma-testnet-7


measure-fi:
## Background

Blanket issue for miscellaneous testnet changes.

## Done when

- https://github.com/palomachain/paloma/pull/302
- https://github.com/palomachain/paloma/pull/303

PR #303, opened 2022-07-05
Increase max validators to 175

measure-fi:
https://github.com/palomachain/paloma/issues/300

## Background

This mirrors the value used in Cosmos

## Testing completed

- [x] existing go tests


Vizualni (approved):
lgtm

PR #369, opened 2022-07-29
Move the addrbook and genesis get after unsafe-reset

taariq:
# Background
Looks like we need to arrange the download of the addrbook AND the genesis AFTER unsafe-reset otherwise, the current flow won't work.

# Testing completed
Just updating the ordering of the README instructions.


Vizualni (approved):
LGTM

PR #375, opened 2022-08-01
Remove version from filename when building the release

Vizualni:
# Background

Taken from: https://goreleaser.com/customization/archive/.
That way version is only written in the release URL which is generate by the github, so that we don't need to write it two times. Also it works with the latest release.

# Testing completed

- [ ] none


measure-fi (approved):
🎉

PR #381, opened 2022-08-03
Use linter proposed by github

Vizualni:
# Related Github tickets

# Background

# Testing completed

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,29 @@
+# This workflow executes several linters on changed files based on languages used in your code base whenever
+# you push a code or open a pull request.
+#
+# You can adjust the behavior by modifying this file.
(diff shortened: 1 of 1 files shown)


measure-fi:
[GitHub Super-Linter](https://github.com/marketplace/actions/super-linter)

Did you not want the `[GitHub Super-Linter](https://github.com/marketplace/actions/super-linter)` badge in the readme? Also does it make sense to add this as an entirely separate workflow (and thus separate container execution) or can it be added to `.github/workflows/ci-test.yml`?

measure-fi:
[GitHub Super-Linter](https://github.com/marketplace/actions/super-linter)

Vizualni:
Reverting this as linter fails to properly load Go modules, thus reports that thing from other packages do not exist.

PR #407, opened 2022-08-16
Update README.md

Vizualni:
# Related Github tickets

# Background

# Testing completed

# Breaking changes


taariq (approved):
Yeup. This is what we need. We should also avoid some of the AppHash errors as well.

verabehr (approved):
lgtm

PR #409, opened 2022-08-17
Pigeon healthcheck

Vizualni:
# Background

We want to stop folks from running paloma chain without them running pigeons. If paloma detects that pigeon is not running, it will kill itself thus stopping a _pigeonless_ validator from entering the valset.

# Testing completed

- [ ] manually tested with both paloma and pigeon

# Breaking changes

Nope. It does not change anything in the state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,7 @@ RestartSec=5
 WorkingDirectory=~
 ExecStartPre=
 ExecStart=/usr/local/bin/palomad start
+Environment=PIGEON_LISTEN_PORT=5757
 ExecReload=
 
 [Install]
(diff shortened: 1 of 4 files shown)


measure-fi:
I can't quite tell, does this only check for a pigeon if we are staking? It is reasonable to run palomad to watch the chain and query it locally without running a pigeon.

Vizualni:
> I can't quite tell, does this only check for a pigeon if we are staking? It is reasonable to run palomad to watch the chain and query it locally without running a pigeon.

Right now it does not. This is only a basic skeleton. From inside of a running chain, you really don't know if you're a validator or a full node, so we can't really do the thing you are asking. We can say that even if you're a simple node, that you must run pigeon.

> I'm also wondering, if they are going to be this tightly coupled, should they just be one process?

Nope. This is something that I initially wanted to do, but it's impossible because cosmos-sdk is intended to be used with only a single chain. This was tested when we wanted to go to cosmos, but we are going to evm instead. We might also go to cosmos as well, so it wouldn't be a good idea to put pigeon in palomad.


----

Also, this is the same technique that other similar chains are using.

PR #411, opened 2022-08-18
Disable logic calls

Vizualni:
# Related Github tickets

# Background

# Testing completed

# Breaking changes

### DIFF
--- a/x/evm/keeper/keeper.go
+++ b/x/evm/keeper/keeper.go
@@ -117,6 +117,9 @@ func (k Keeper) AddSmartContractExecutionToConsensus(
 	turnstoneID string,
 	logicCall *types.SubmitLogicCall,
 ) error {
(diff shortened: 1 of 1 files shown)


x/evm/keeper/keeper.go
```diff
@@ -117,6 +117,9 @@ func (k Keeper) AddSmartContractExecutionToConsensus(
 	turnstoneID string,
 	logicCall *types.SubmitLogicCall,
 ) error {
+	k.Logger(ctx).Info("wanting to add smart contract execution to chain. skipping for now as it's not yet supported officially")
+	return nil
+
 	return k.ConsensusKeeper.PutMessageInQueue(
```
measure-fi, line 123: Does go not warn this line is unreachable 🤨

Vizualni: I had to disable linter

Issue #283, opened 2022-06-29
Move all those Welcome-*.md files from root of the project into docs folder


Vizualni:
# Background

We have a bunch of files at the root of the project which are cluttering the precious root directory.
We should move those in the docs folder and create links to those welcome files under the main README.me.

# Done when

- [ ] welcome files are moved to docs folder
- [ ] links pointing to those welcome files are created in README.md

PR #436, opened 2022-08-24
Welcome files

verabehr:
# Related Github tickets

https://github.com/palomachain/paloma/issues/283


taariq (approved):
These look good. However, tomorrow we are moving to paloma-testnet-9 with the new upgrade so not sure that this PR keeps up.

Vizualni (approved):

PR #460, opened 2022-09-01
libwasmvm.x86_64.so path has changed at cosmwasm repo

mahofmahof:
# Related Github tickets

# Background

# Testing completed

# Breaking changes


measure-fi:
Thanks for the PR ^_^

We should probably pull this library into paloma, or pin that link to a particular commit. I think @Vizualni expected this workflow to be more temporary.

Issue #465, opened 2022-09-09
Egg minting contract - payload as byte array


verabehr:
# Background

https://www.notion.so/volumefi/EPIC-09-16-Paloma-Handles-Target-Chain-Logic-Call-Construction-from-Cosmwasm-Contracts-2b070ab0762548df82d80aed69db93be

# Done when

- https://github.com/palomachain/paloma/issues/464
- [ ] we can mint an egg

Issue #464, opened 2022-09-09
Payload data as byte array


verabehr:
# Background
https://www.notion.so/volumefi/EPIC-09-16-Paloma-Handles-Target-Chain-Logic-Call-Construction-from-Cosmwasm-Contracts-2b070ab0762548df82d80aed69db93be

Relevant paloma-rs issue: https://github.com/palomachain/paloma/issues/465

# Done when

- [ ] payload data is not empty and sent in the form the cosmwasm developer constructs it

measure-fi:
@Vizualni Was wondering if you have a good idea of exactly the format you would like to use here. Happy to make up multiple contracts to test.

PR #466, opened 2022-09-12
Accept payload as byte array

Vizualni:
# Related Github tickets

- #465 
- #464 

# Background

Changes the input type for the payload variable from string to byte. Serializing string as byte[] is a bit clumsy.

# Testing completed

# Breaking changes

### DIFF
--- a/x/evm/keeper/keeper.go
+++ b/x/evm/keeper/keeper.go
@@ -310,13 +310,15 @@ type executeEVMFromCosmWasm struct {
 		CompassID string `json:"compass_id"`
 	} `json:"target_contract_info"`
 
-	// TODO: we need to have this as a payload
-	Payload string `json:"payload"`
+	Payload []byte `json:"payload"`
 }
 
 func (e executeEVMFromCosmWasm) valid() bool {
 	zero := executeEVMFromCosmWasm{}
-	if e == zero {
+	if e.TargetContractInfo == zero.TargetContractInfo {
+		return false
+	}
+	if len(e.Payload) == 0 {
 		return false
 	}
 	// todo: add more in the future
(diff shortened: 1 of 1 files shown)


taariq (approved):
LGTM!

x/evm/keeper/keeper.go
```diff
@@ -310,13 +310,15 @@ type executeEVMFromCosmWasm struct {
 		CompassID string `json:"compass_id"`
 	} `json:"target_contract_info"`
 
-	// TODO: we need to have this as a payload
-	Payload string `json:"payload"`
+	Payload []byte `json:"payload"`
```
measure-fi, line 313: Does this mean it will be an array like `[92,183,148,210,0,134,193,229,48,158,88,192...]`? Because there are possibly more efficient encodings.

Vizualni: Nope. Golang's json de/encoder does that with base64. How is rust handling this?

measure-fi: Base 64. Just wondering because the comment on this function implies it might be the array format.

Vizualni: Good point. Now that you're pointing this out to me, it seems that I have put string there for a reason because when I was testing things that array thingy for the eth_address caught my attention as well. Could you please write a super simple smart contract that emits the event for executing an EVM from and simply use payload as a byte array? Basically, to use the same types that you'd use when preparing EVM call.

Vizualni:
> Does this also mean the contract will need to encode everything appropriately for the receiver chain?

Yes. BTW just to mention this once again, this way of executing jobs is a temporary one. Once I start my work on gas management & scheduler, you'd only need to pass in the job_id and payload.

x/evm/keeper/keeper.go
measure-fi: https://github.com/palomachain/paloma-rs/pull/47

PR #516, opened 2022-10-03
Disable paloma if the version is different than what the government said should be

Vizualni:
# Background

Disable paloma if the version is different than what the government said should be
# Testing completed

- [x] wrote tests

# Breaking changes
Not a breaking change.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -427,7 +427,9 @@ func New(
 		keys[palomamoduletypes.StoreKey],
 		keys[palomamoduletypes.MemStoreKey],
 		app.GetSubspace(palomamoduletypes.ModuleName),
+		semverVersion,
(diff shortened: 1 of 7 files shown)


taariq:
"the government"? We mean "governance". However, why would paloma be disabled if someone was running a lower version?

Vizualni:
> "the government"? We mean "governance". However, why would paloma be disabled if someone was running a lower version?

Yeah. It's the governance. My bad :). This is simply an alarm that will let people know that they are not running the correct paloma version. This check stops those who are not going to use cosmovisor for upgrades. Paloma chain service will be stopped, but using paloma client is still allowed. This is also to trying to eliminate the AppHash mismatch errors due to people running different paloma versions.

Issue #526, opened 2022-10-05
Add Codeql analyzer to CI workflow


verabehr:
# Background

_Describe the problem that we are trying to solve with this feature._
_If a path to a solution is known, feel free to write it down, but have an open mind_

Alex from Confio suggested that adding https://github.com/CosmWasm/wasmd/blob/main/.github/workflows/codeql-analizer.yml will make our lives easier.

# Done when

- [ ] https://github.com/CosmWasm/wasmd/blob/main/.github/workflows/codeql-analizer.yml was added to Paloma's CI

taariq:
Let's make our Cosmwasm debugging lives easier!

taariq:
Closing as distraction to gas management.

PR #528, opened 2022-10-05
Add codeql-analyzer to CI

measure-fi:
https://github.com/palomachain/paloma/issues/526

## Background

This analyzer may be a little noisy initially, but will hopefully tell us where new issues are creeping in. Going to silence the remainder of the warnings and merge it (for now).

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -1,7 +1,4 @@
 on:
-  push:
-    branches:
-      - '*'
   pull_request:
     branches:
       - '*'
--- a/.github/workflows/codeql-analyzer.yml
+++ b/.github/workflows/codeql-analyzer.yml
@@ -1,0 +1,35 @@
+name: "Code Scanning - Action"
+
+on:
+  pull_request:
+    branches:
+      - '*'
+  push:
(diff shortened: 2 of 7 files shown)


taariq:
I love curiosity! Let's be curious!

.github/workflows/ci-test.yml
```diff
@@ -1,7 +1,4 @@
 on:
-  push:
-    branches:
-      - '*'
```
measure-fi, line 4: @Vizualni Without this it's just running CI twice against every PR. If you want CI running on all branches I would turn it off for PRs.

taariq:
Closing for now.

measure-fi:
@taariq One of these errors did line up with the fixes last week. I think the payoff here is rather large: it avoids a chain halt. I'd budget 1.5 days of my time to fix all of these and then leave it on in CI to audit future work. When exactly that happens is a matter of prioritization, but it should happen.

taariq:
@measure-fi re-opening!

measure-fi:
I think the floating point error is _actually_ bad, so I'm leaving it in for now. Will need more thought on that one.

PR #550, opened 2022-10-24
update readme to v11.1

verabehr:


taariq (approved):
Yeup. Tested in production and it works.

PR #558, opened 2022-10-27
Fix FromMapKeys, resolve some go map iteration issues.

measure-fi:
## Background

It's important on the critical path to iterate maps in order, this function will help accomplish that.

## Testing completed

- [ ] Existing tests pass.

### DIFF
--- a/util/slice/go_map.go
+++ b/util/slice/go_map.go
@@ -2,21 +2,24 @@ package slice
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


taariq:
@measure-fi where is the ticket for this update? I'm not sure what this feature does and why it was merged.

measure-fi:
@taariq https://github.com/palomachain/paloma/issues/526, added to description.

Issue #593, opened 2022-12-23
Upgrade paloma's keyring backend from "test" to "os" to align with Pigeon


taariq:
# Background
Upgrade paloma's keyring backend from "test" to "os" to align with Pigeon. Currently paloma's keyring back-end is "test" and we will want to upgrade to os.

# Done when

- [ ] patch and upgrade keyring storage on paloma to match pigeon.

PR #595, opened 2022-12-26
refactor: change default keyring backend

alexanderbez:
## Changelog

- Change default keyring backend from `test` to `os`

closes: #593


taariq (approved):
We close the keyring backend issues with this change.

PR #616, opened 2023-01-03
update logs

webelf101:
# Related Github tickets

# Background

# Testing completed

# Breaking changes


verabehr (approved):

taariq:
@verabehr & @webelf101 we'll need to remove the printfs for this PR so that we don't make the logs overly noisy on master.

verabehr:
Is it possible to keep them and change the level  to "debug" instead? They
might be useful in the future.

taariq:
No. Those are print statements, no logging. It will print out those messages on the server logs.

taariq:
@webelf101 am I correct here?

PR #617, opened 2023-01-03
add new relic

rootedbox:
# Related Github tickets

# Background

# Testing completed

# Breaking changes


taariq:
Closing. we don't want a PR on master for this.

go.mod
```diff
@@ -3,10 +3,9 @@ module github.com/palomachain/paloma
 go 1.19
 
 require (
-	github.com/CosmWasm/wasmd v0.30.0
```
rootedbox, line 6: so revert these changes the `go get github.com/newrelic/go-agent/v3` did? @verabehr

verabehr: Is it required to downgrade? If so, we may need to look further for an alternative solution, but if it is not required, we definitely don't want to downgrade any of these libraries. They're all core libraries.

PR #618, opened 2023-01-04
refactor: change stringer for QueuedSignedMessage

alexanderbez:
## Changelog

* Remove the default Proto stringer for `QueuedSignedMessage`
* Implement a custom `String` method for `QueuedSignedMessage`

I still don't know why the panic was being triggered on this type, so consider this patch, if it even works, to be a duct tape solution.


verabehr (approved):
lg and works

PR #647, opened 2023-01-11
chore: update contrib doc

alexanderbez:


taariq (approved):
Yup. This the stuff. Thank you!

PR #670, opened 2023-01-20
chore: update release GH workflow

alexanderbez:


.github/workflows/release.yml
```diff
@@ -1,42 +1,31 @@
...
-      - name: Set up QEMU
-        id: qemu
-        uses: docker/setup-qemu-action@v2
-
-      -
-        name: Use Docker Buildx
-        id: buildx
-        uses: docker/setup-buildx-action@v2
-      -
-        name: Build image for release
```
taariq, line 30: Whats the impact of removing the images for release here?

.github/workflows/release.yml
```diff
@@ -1,42 +1,31 @@
 name: Release
-
+# This workflow helps with creating releases.
+# This job will only be triggered when a tag (vX.X.x) is pushed
 on:
   push:
+    # Sequence of patterns matched against refs/tags
     tags:
-      - '*'
+      - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
```
taariq, line 8: Yeup. This was sorely needed. Thank you for adding.

.github/workflows/release.yml
alexanderbez: This job uses go-releaser directly to build and publish releases. Not sure why Docker was used...

taariq (approved):
This is the refactor that we needed. Thank you!

PR #671, opened 2023-01-20
add testnet-13 validators

verabehr:


taariq (approved):
Thank you for the update!

PR #683, opened 2023-01-27
Update .mergify.yml

alexanderbez:
Update Mergify config to include `release/v1.x.x` backport label.


taariq (commented):
LGTM! Semver compliance here we come!

taariq (approved):

PR #716, opened 2023-02-13
add log

webelf101:
# Related Github tickets

# Background

# Testing completed

# Breaking changes


taariq (approved):
Approving. Question: Is this debug or INFO?

mszuyev:
@taariq none of them, you lint the code before actual deploying it. You can do it locally (as a developer) or in a github-ci worker (CI/CD). So linting message is different from application logs.
When app is built, there will be nothing more to lint (since no source .go files with the binary are deployed)

PR #758, opened 2023-03-06
ignore cosmos in dependabot

mszuyev:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/6

# Background

Cosmos SDK updates are backward incompatible, we want to update it manually


taariq (approved):
LGTM!

PR #791, opened 2023-04-08
chore: update upgrade handler

alexanderbez:
# Related Github tickets

# Background

# Testing completed

# Breaking changes


alexanderbez:
I'm not sure if we need to use ibc-go v6 migration as we're coming from v4 on mainnet. However, it's not possible to import ibc-go v6 using Cosmos SDK v0.47

I guess trying this out on a testnet will show us.

taariq (approved):
LGTM!

Issue #811, opened 2023-04-27
Create MemStoreKeys for custom modules


0xmuralik:
MemStoreKeys must be created first using the `sdk.NewMemoryStoreKeys` [here](https://github.com/palomachain/paloma/blob/f2e6d757e296e89c09e4456286b49bb06a71524c/app/app.go#L332). And then must be accessed through a different collection say `memKeys` instead of the [current appraoch](https://github.com/palomachain/paloma/blob/f2e6d757e296e89c09e4456286b49bb06a71524c/app/app.go#L507).

PR #814, opened 2023-04-27
create mem store keys

0xmuralik:
# Related Github tickets

- Closes: #811 

# Background

# Testing completed

# Breaking changes

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -184,7 +184,6 @@ var (
 		ibctm.AppModuleBasic{},
 		ica.AppModuleBasic{},
 		ibcfee.AppModuleBasic{},
-		consensus.AppModuleBasic{},
 	)
 
 	// module account permissions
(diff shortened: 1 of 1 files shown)


app/app.go
```diff
@@ -184,7 +184,6 @@ var (
 		ibctm.AppModuleBasic{},
 		ica.AppModuleBasic{},
 		ibcfee.AppModuleBasic{},
-		consensus.AppModuleBasic{},
```
0xmuralik, line 187: registered twice

taariq (approved):

app/app.go
taariq: Is that a problem?

0xmuralik: https://github.com/palomachain/paloma/blob/f2e6d757e296e89c09e4456286b49bb06a71524c/app/app.go#L181
https://github.com/palomachain/paloma/blob/f2e6d757e296e89c09e4456286b49bb06a71524c/app/app.go#L187
It's not required to add it twice, right?

taariq: No. It's not. Let's remove one so we keep only one.

Issue #816, opened 2023-04-27
Pigeon Bug: Unknown desc = can't route message creator unknown Request while trying to keep pigeon alive


taariq:
# What is happening?
<details>
  <summary>Section description</summary>
<i>Provide as much context as you can. Give as much context as you can to make it easier for the developers to figure what is happening.</i>
</details>

It is unclear which rpc command is causing this error, but it appears to happen with the keep alive messages. 

{"error":"rpc error: code = Unknown desc = rpc error: code = Unknown desc = can't route message creator:\"paloma1s3p2xfe65mwcjdxn79nfg2rythlya28vtf08tf\" : unknown request [cosmos/cosmos-sdk@v0.47.1/baseapp/baseapp.go:786] With gas wanted: '18446744073709551615' and gas used: '27066' : unknown request","level":"error","msg":"error while trying to keep pigeon alive","time":"2023-04-27T13:44:59Z"}

It appears that a message handler is missing for Keep Alive messages which was not a problem before in v0.11.7

# Paloma and pigeon versions and logs
<details>
  <summary>Section description</summary>
<i>Write down paloma version.</i>
<i>Write down pigeon version.</i>
<i>Copy and paste pigeon config file as well as relevant ENV variables.</i>
</details>

1. Paloma Master
2. Pigeon Master

# How to reproduce?
<details>
  <summary>Section description</summary>
  <i>Please write detailed steps of what you were doing for this bug to appear.</i>
</details>

Run 3-node network and run pigeon.

# What is the expected behaviour?
<details>
  <summary>Section description</summary>
<i>If you know, please write down what is the expected behaviour. If you don't know, that's ok. We can have a discussion in comments.</i>
</details>

Keep alive messages work as intended.

taariq:
UPDATE: Looks like the app doesn’t know how to route the message which is the Keep Alive Message.

taariq:
Closing this ticket as it appears it is related to upgrading from deprecated MessageHandlers from pre cosmos-SDK v0.45.

PR #818, opened 2023-04-28
Upgrade the way of handling messages

ToasterTheBrave:
# Related Github tickets
* https://github.com/palomachain/paloma/issues/816
* 

# Background

It was pointed out to us that the old way of handling messages (which we were doing with the handler.go) is no longer supported in cosmos-sdk v0.47.1  We have to update.  This does that.

# Testing completed

- [x] This was tested on a local private testnet and two things were observed that indicate this works as expected
  * The error message from pigeon have stopped
  * The nodes are no longer being jailed

# Breaking changes


taariq (approved):
Nice! Thank you @ToasterTheBrave. This is what we needed. Thank you!

PR #824, opened 2023-05-04
Setting the Tx Encoded in the app

ToasterTheBrave:
This solves https://github.com/VolumeFi/paloma/issues/148

# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/148

# Background

# Testing completed

# Breaking changes


taariq (approved):
one line!

PR #830, opened 2023-05-05
Tweak matching string because it's not regex

ToasterTheBrave:


.github/workflows/release.yml
```diff
@@ -5,7 +5,7 @@ on:
   push:
     # Sequence of patterns matched against refs/tags
     tags:
-      - "v[0-9]+.[0-9]+.[0-9]+.*" # Push events to matching v*, i.e. v1.0, v20.15.10
+      - "v[0-9]+.[0-9]+.[0-9]+-?*" # Push events to matching examples like v1.2.3, v1.2.3-beta1
```
measure-fi, line 8: Explicitly allowing `v1.2.3-`? Maybe the language doesn't let you do better :-/

measure-fi: Also note that `*` does not match `/`, but `**` would if that was your intent.

measure-fi (approved):

.github/workflows/release.yml
measure-fi: Actually another thing, I guess you wanted to record the intent to do something like `v1.2.3-beta`, but this still let's me exclude the `-`. I think _maybe_ this is most accurately
```
tags:
  - "v[0-9]+.[0-9]+.[0-9]+"
  - "v[0-9]+.[0-9]+.[0-9]+-**"
```
_but I'm not sure_

measure-fi: But maybe `v[0-9]*` would be plenty.

PR #833, opened 2023-05-09
Fix decoding of wasm job payload on execution

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/169
- https://github.com/VolumeFi/paloma/issues/166

# Background

Executing wasm contracts appears to never have worked.  The executor for these jobs was incorrectly decoding the job payload

# Testing completed

- [x] I added a unit test to ensure the decoding happens correctly

# Breaking changes


taariq (approved):
So UnmarshallJob function added and then execute the payload. Got it. Looking forward to learning more.

taariq:
Just a failing lint check.

PR #847, opened 2023-05-18
Add in glibc gotcha and example to build from source

ToasterTheBrave:
# Background

The crux of this issue is that the version of glibc on the github machine that builds this vs that of those who run it may be different. Until this is fixed, we need to add instructions for people to build this on their own machine

# Testing completed

- [x] Ran through the build steps locally

# Breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,26 +58,33 @@ See [Release procedure](CONTRIBUTING.md#release-procedure) for more information 
 
 ## Join an active Network
 
-To get the latest `palomad` binary:
+> #### Note:
+> Some have seen errors with GLIBC version differences with the downloaded binaries for v1.0.0.  This is caused by a difference in the libraries of the host that built the binary and the host running the binary.
+>
+> We did not find this bug in time to fix it in the v1.0.0 release.  Instead there is a workaround.  If you experience these errors, please pull down the code and build it, rather than downloading the prebuilt binary
 
-Testnet:
+### To get the latest prebuilt `palomad` binary:
+
 ```shell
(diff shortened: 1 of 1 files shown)


measure-fi (approved):
Is there a way to vendor glibc? Or link against something like musl?

ToasterTheBrave:
Yes, I that's what the CGO_ENABLED=0 flag basically does.  Problem is, that threw some errors in the cosmwasm module.  We can probably achieve this with specific flags during the build, but it seems too risky to put into an already tested release.  It should get thoroughly tested before being released.

README.md
```diff
@@ -58,27 +58,34 @@ See [Release procedure](CONTRIBUTING.md#release-procedure) for more information
...
+> We did not find this bug in time to fix it in the v1.0.0 release.  Instead there is a workaround.  If you experience these errors, please pull down the code and build it, rather than downloading the prebuilt binary
+
+### To get the latest prebuilt `palomad` binary:
 
-Testnet:
 ```shell
-wget -O - https://github.com/palomachain/paloma/releases/download/v0.11.7/paloma_Linux_x86_64.tar.gz  | \
-  sudo tar -C /usr/local/bin -xvzf - palomad
+wget -O - https://github.com/palomachain/paloma/releases/download/v1.0.0/paloma_Linux_x86_64.tar.gz  | \
+  sudo tar -C ~/ -xvzf - palomad
```
taariq, line 70: This doesn't extract to /usr/local/bin anymore, but to user home directory. Any reason why we are making this change?

PR #848, opened 2023-05-19
Revert building with CGO disabled

ToasterTheBrave:
# Background

This was done to try and get around glibc errors related to building on one machine and running on another.  This caused unexpected errors elsewhere and needs reverted to keep the master branch buildable.


taariq (approved):

taariq:
@ToasterTheBrave approved, but I'll let you merge.

PR #849, opened 2023-05-19
Run go mod tidy

ToasterTheBrave:
# Background

After the recent updates from dependabot, we need to tidy the go.mod file


taariq (approved):
LGTM!

PR #857, opened 2023-05-25
Sender public key now being sent with scheduled jobs

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/243

# Background

We want to allow contracts to verify that they were called by an entity with the desired permissions to run the contract.  To do this, we need to pass in an identifier of the caller alongside the payload.
    
There has been some back and forth on this issue.  The current implementation is to send the "sender" address along with the payload to the contract.


measure-fi:
Is the executor the public key that is executing the contract or (in the case that the contract is executing a job) the public key of the contract, or some third thing?

ToasterTheBrave:
It's the public key of the account that submitted the execute job command in paloma.

ToasterTheBrave:
Maybe we should use a different name than "executor" then.  We could do "sender"

measure-fi:
So `sender` is used within the contract as the address that triggered execution. But we may want to consider the job as being kicked off by the address of the contract. One of the points of a contract is that I might have any number of people allowed to execute it (and/or other contracts), but I actually want to grant only that contract as allowed to kick off a job.

(I think)

verabehr:
> One of the points of a contract is that I might have any number of people allowed to execute it (and/or other contracts), but I actually want to grant only that contract as allowed to kick off a job.

yes, that is my understanding as well

ToasterTheBrave:
Converting this back to draft.  It appears after some further discussion that we prefer the first implementation, where the sender's identifier is passed rather than the contract's identifier.

ToasterTheBrave:
This is now ready for review after reverting back to the original implementation

taariq (approved):
Great update. Let's get this into private testnet for testing.

PR #858, opened 2023-05-26
Send different identifier for cosmwasm executed jobs

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/243

# Background

We want a different identifier based on who is submitting the job.  If a user submits the job, we want the user's public key.  If a job is submitted through a cosmwasm contract, we want the contract address.


taariq (approved):
Okay. We now cover both Go and Coswasm/Rust sides.

PR #859, opened 2023-05-26
Remove the go mod tidy command from protogen script

ToasterTheBrave:
# Background

There is no reason to run a mod tidy after generating the proto files. This just adds unnecessary time every time we need to generate these. Removing it.


taariq (approved):

measure-fi:
@ToasterTheBrave

I agree that `go mod tidy` is probably bad here, and no reason is given for adding it on cosmos-sdk either. The last time I ran this script I got a load of deprecation errors. I wonder if it isn't a good idea to adopt the [upstream changes](https://github.com/cosmos/cosmos-sdk/blob/main/scripts/protocgen.sh) wholesale, or if that removes any warnings.

By the way, is CI running this script, checking that it doesn't change any files?

PR #866, opened 2023-05-30
Hotfix/v1.1.0 upgrade fix

ToasterTheBrave:
# Background

When starting up, paloma is attempting to run upgrades meant for a previous version (v1.0.0).  This code should not be dynamic, as all upgrades are different.  We are now specifying the exact version this code is meant to run for.

# Testing completed

- [x] This was tested on a local private testnet


ToasterTheBrave:
Don't merge this yet.  It accidentally has some other PRs from the release that I need to strip out.  I'll re-open a new PR

PR #868, opened 2023-05-30
Fixing broken v1.1.0 version

ToasterTheBrave:
# Background

When starting up, paloma is attempting to run upgrades meant for a previous version (v1.0.0).  This code should not be dynamic, as all upgrades are different.  We are now specifying the exact version this code is meant to run for.

# Testing completed

- [x] This was tested in a local private testnet


taariq (approved):
LGTM

PR #871, opened 2023-06-08
Just-in-time valset updates

ToasterTheBrave:
Closes https://github.com/VolumeFi/paloma/issues/273

# Background

We don't want to send valset updates every 50 blocks for chains that are relatively inactive.  This is a waste of money and resources.  Instead, we are switching to just-in-time valset updates.  

Snapshots are still taken every 50 block if worth, they just aren't pushed to target chains on every snapshot.  Instead, valsets are now published to a chain in one of two cases:
1. If it's been a week since the last published valset on a chain
2. If a job is executed and the most recent valset has not been published to the chain yet.

# Testing completed

- [x] wrote tests
- [x] tested manually on a private testnet

# Breaking changes


x/evm/keeper/attest_test.go
```diff
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"errors"
```
taariq, line 4: This is helpful.

taariq (approved):
Okay. We have tests and now snapshots that are less inefficient.

PR #875, opened 2023-06-09
Adding some mocks needed by pigeon

ToasterTheBrave:
# Background

The pigeon repo is currently vendoring these mocks and shouldn't be.  The mocks belong in the paloma repo to prevent them becoming out of date, which is already the case in Pigeon.  This work is necessary as a pre-step to the next pigeon feature.


taariq (approved):
get them mocks!

PR #876, opened 2023-06-13
Add error data to messages in turnstone queue

ToasterTheBrave:
# Related Github tickets

# Background

This is needed to change Pigeon behavior.  Currently, when a Pigeon encounters an error relaying a message, the pigeon just attests the message and then won't touch it again.  This means for a bad message, the majority of pigeons must try to relay it in order to achieve enough attestations.

With this new functionality, when an Pigeon encounters an error, it can send that error data back to Paloma.  That way, all other pigeons can attest to the error without having to attempt to relay it themselves.

# Testing completed

- [x] This was tested on a local private testnet
- [x] Tests were added

# Breaking changes


taariq (approved):
To reflect: We now get error telemetry attached to messages when pigeons relay and fail.

ToasterTheBrave:
> To reflect: We now get error telemetry attached to messages when pigeons relay and fail.

That is correct

PR #877, opened 2023-06-14
Update README.md

verabehr:


measure-fi (approved):

measure-fi:
> Update README.md

Typically it would be nice if the commit message/pr title were meaningful to the change, this isn't really updating so much as fixing the header for a shell block in the markdown of the readme.

PR #879, opened 2023-06-16
Fix payload sender appending

ToasterTheBrave:
Closes https://github.com/VolumeFi/paloma/issues/311

# Background

This feature was almost complete, but it was never QA tested.  Now that we're testing, it needs some modifications
* Actually append the sender data to the payload at this level
* Send 32 bytes, 0 padded of the hex encoded sender or contract address

# Testing completed

- [x] tested in a private tesetnet
- [x] wrote tests

# Breaking changes


taariq (approved):
LGTM and test some more!

internal/x-chain/types.go
```diff
@@ -35,7 +35,7 @@ type CobraTXJobAdder interface {
 
 type Jobber interface {
 	VerifyJob(ctx sdk.Context, definition []byte, payload []byte, refID ReferenceID) (err error)
-	ExecuteJob(ctx sdk.Context, definition []byte, payload []byte, senderPubKey []byte, contractAddress []byte, refID ReferenceID) (err error)
+	ExecuteJob(ctx sdk.Context, definition []byte, payload []byte, senderAddress sdk.AccAddress, contractAddress sdk.AccAddress, refID ReferenceID) (err error)
```
measure-fi, line 38: Yay.

x/evm/keeper/scheduler_job.go
```diff
@@ -63,17 +66,55 @@ func (k Keeper) ExecuteJob(ctx sdk.Context, definition, payload []byte, senderPu
 	if err != nil {
 		return err
 	}
+
+	var hexBytes []byte
+	switch {
+	case senderAddress != nil:
+		hexBytes, err = hex.DecodeString(addressToHex(senderAddress))
```
measure-fi, line 73: What is this pair of methods doing because it looks like it just does `unhex(hex(blob))`. Why does that not mean `hexBytes := blob`?

PR #882, opened 2023-06-19
Toaster/module updates

ToasterTheBrave:
# Background

Updating all dependabot mentioned dependencies that are drop in replacements without additional work

# Testing completed

- [x] ran tests


taariq (approved):
Upgraded and rollin!

PR #889, opened 2023-06-20
Slow down rejailing slightly

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/254

# Background

This is meant as a stop-gap until we get to a better solution.  Pigeons are rejailed too quickly when unjailing.  We're slowing this process down slightly by not checking quite as often

# Testing completed

# Breaking changes


taariq (approved):
LGTM!

PR #891, opened 2023-06-23
Adding commands for managing snapshots

ToasterTheBrave:
# Background

This was done as part of a debugging effort, but seems useful to have in palomad.  Using these commands, we can easily manage snapshots ad-hoc

# Testing completed

- [x] Tested on testnet during debugging

# Breaking changes


taariq (approved):

taariq (approved):
👍

PR #892, opened 2023-06-23
Adding fee management through governance

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/341

# Background

We have two fees we want for the gas management epic, and we want them managed through governance.  This PR achieves that.  This introduces two new gov commands: `propose-community-fund-fee` and `propose-security-fee`.  These are kept as separate commands so as to not create a case where you agree with only half of a proposal.

There is a little bit of refactoring that's going to be an ongoing work in progress to allow better unit testing of our code.

# Testing completed

- [x] wrote tests
- [x] wrote ansible tests for our private testnet https://github.com/palomachain/private-testnet/pull/30
- [x] tested on a local private testnet

# Breaking changes


taariq (approved):
Reviewed and these are for governance.

PR #896, opened 2023-07-03
Don't jail when starting up a new chain

ToasterTheBrave:
# Background

This really just helps development.  We want to give both Paloma and Pigeon some time before we start worrying about jailing.  Now, we won't jail within the first 50 blocks of a new chain

# Testing completed


taariq (approved):
LGTM!

PR #897, opened 2023-07-04
🔧 fix: #265 remove yaml marshalling in endblock logging

byte-bandit:
# Related Github tickets

https://github.com/VolumeFi/paloma/issues/265

# Background

During the endblock handling, the entire message would get logged. The [stringer implementation](https://pkg.go.dev/golang.org/x/tools/cmd/stringer) for the underlaying message type was using the yaml package for a complete string marshalling, which resulted in a large amount of overhead. 

Logging the entire message in byte code not only take quite a long time due to marshalling, it's also largely irrelevant, as message information can easily be obtained from the queue.

# Testing completed

- [x] _a list of actions that you've performed to ensure that this PR works as intended_

# Breaking changes

### DIFF
--- a/x/consensus/keeper/attest.go
+++ b/x/consensus/keeper/attest.go
@@ -1,8 +1,6 @@
 package keeper
 
 import (
-	"fmt"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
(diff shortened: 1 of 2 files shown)


taariq (approved):
Let's get it out!

PR #898, opened 2023-07-05
feat: #254 Jail pigeons after timeout

byte-bandit:
Pigeons will no longer be jailed at a predefined block height. Instead, a fixed timout is used. By default, pigeons will reset their connection TTL by 5 minutes every few seconds, as well as with every message call.

If for some reason, the validator fails to connect, pigeons become eligible for jailing, which happens once per minute. After five minutes without a heart beat, pigeons become jailed, with a warning error message being logged up to 30 seconds in advance.

Pigeons may reconnect while jailed, but will not be in any validator sets until unjailed and rebonded.

# Related Github tickets

See [#254](https://github.com/VolumeFi/paloma/issues/254)

# Testing completed

# Breaking changes


taariq (approved):
LGTM!

PR #899, opened 2023-07-05
🔧 fix: ensure jail warning is logged

byte-bandit:
With jailing rounds happening once per minute, a 30 second threshold before TTL runs out might not be printed in time for users to react.

Switching the threshold to two minutes will give them ample to time to investigate the issue and potentially restart pigoen.

# Testing completed

- [x] _a list of actions that you've performed to ensure that this PR works as intended_
- [x] _99% of the times you should have an item: "wrote tests"_

# Breaking changes


byte-bandit:
This is closed with https://github.com/palomachain/paloma/pull/909

PR #902, opened 2023-07-05
Revert "feat: #254 Jail pigeons after timeout"

ToasterTheBrave:
Reverts palomachain/paloma#898

This appears to need some more work (and rolled back until the work is done).  It's causing errors when I run it in my local testnet, and most certainly will in mainnet

If I just kill a pigeon and let it reach the end of the TTL, palomad throws an error and breaks the chain.
```
Jul 05 18:50:31 m2 palomad[78744]: {"level":"error","module":"server","module":"consensus","height":217,"round":6,"err":"wrong Block.Header.AppHash.  Expected 437036D053C17E21AF21717F13B7479982589A423E68D757B8C996D6A2EEDC2F, got DBC723F7850C095868326648BFDE22D9C9AC26205190F5DBC48316DD84A00998","time":"2023-07-05T18:50:31Z","message":"prevote step: consensus deems this block invalid; prevoting nil"}
```

This error means we've created inconsistency between nodes.  The other nodes don't have the same data in their consensus module as this node.

### DIFF
--- a/x/valset/keeper/keep_alive.go
+++ b/x/valset/keeper/keep_alive.go
@@ -14,10 +14,7 @@ import (
 
 const (
 	// TODO: make this a param
-	defaultKeepAliveDuration        = 5 * time.Minute
-	cValidatorJailedErrorMessage    = "validator is jailed"
-	cValidatorNotBondedErrorMessage = "validator is not bonded"
-	cJailingImminentThreshold       = time.Second * 30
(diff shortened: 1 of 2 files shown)


taariq (approved):
Let's revert!

byte-bandit (approved):
Good catch. This means my local build is not running as expected yet.

PR #912, opened 2023-07-13
Ignoring a local directory used for notes

ToasterTheBrave:
# Background

This is a directory commonly used for notes and local data.  Ignoring it in git


byte-bandit (approved):
Perfect!

PR #913, opened 2023-07-13
Delete messages that are stuck in queues

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/461

# Background

We sometimes see messages stuck in our queues and need to address this problem.

Every 10 blocks, check the age of the messages currently in the queues. If a message is over 300 blocks (~30 minutes) old, we consider it stuck and delete it.  This prevents stuck messages from filling our queues.

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/keeper/cleanup.go
+++ b/x/consensus/keeper/cleanup.go
@@ -1,0 +1,30 @@
+package keeper
+
+import (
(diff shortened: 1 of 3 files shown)


byte-bandit (approved):
lgtm

PR #914, opened 2023-07-13
feat: link to op-geth

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/439

# Background

This change introduces `op-geth`, which should support both `bnb-main` and `op-main`.

# Testing completed

# Breaking changes


byte-bandit:
`arb-main` testing is still ongoing, but one way or another, it seems we can move fast towards Optimism at least.

ToasterTheBrave (approved):
Looks good

PR #915, opened 2023-07-13
Better message limit handling

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/461

# Background

This will help with a couple things
1. If a single pigeon is acting up and leaving all of their messages in the queue, other pigeons will still get their messages.  A single pigeon's messages can no longer block other messges
2. If our queue is overfull, we can now see all messages currently in the queue, not just the first 200

Work done:
* Don't always require a limit on messages
  * Adding an option to get all messages, not just N messages.  This is useful for times when we are going to filter the messages downstream.
* Move message response counts
  * If we request a number of messages, we want that to be the number
    returned, not the number before filtering
  * Also fixing places that wanted all messages.  Now that we can return
    all, let's just request all
* Return all messages for requested messages in queue
  * This is used for the cli command

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/keeper/attest.go
+++ b/x/consensus/keeper/attest.go
@@ -14,7 +14,7 @@ func (k Keeper) CheckAndProcessAttestedMessages(ctx sdk.Context) error {
 			return err
 		}
 		for _, opt := range opts {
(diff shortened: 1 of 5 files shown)


byte-bandit (approved):
🙌

PR #919, opened 2023-07-14
Enforce Pigeon minimum version in Paloma

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/462

# Background

This has been a long time coming.  We often release code that is incompatible with previous versions of Pigeon.  Even when done in a backwards-compatible way, backwards compatibility isn't meant to extend backwards forever.  We need a way of specifying the minimum version of Pigeon that the current version of Paloma works with.  It's currently expected that any newer versions of Pigeon will also work with the current version of Paloma

The minimum required version of Pigeon is specified in the app.go file and should be mentioned in any proposed paloma update that goes through governance.  Users can upgrade to the latest version of Pigeon ahead of the Paloma upgrade.

Part of this PR includes cleaning up extra places where we were calling the keepalive functionality.  Keepalive should only happen in its own workflow, not as a tag-along to other workflows.

I also removed the ability to send a keepalive message with the CLI.  I can come up with no reason why we would want to have this ability.

More info can be seen in the Epic: https://www.notion.so/volumefi/Epic-07-14-23-Governance-for-pigeon-version-check-8f8fd1dbd4d24a40a3007294f172c738

# Testing completed

# Breaking changes


app/app.go
```diff
@@ -140,6 +140,8 @@ const (
 	Name = "paloma"
 
 	wasmAvailableCapabilities = "iterator,staking,stargate,paloma"
+
+	minimumPigeonVersion = "v1.4.0"
```
taariq, line 144: If this is set here, how is it changed at governance for the upgrade to say pigeon `v.1.5.0`

ToasterTheBrave: When we have a new version of Paloma, there are two possibilities:
1. The new version is still compatible with all Pigeon versions down to v1.4.0
    * No Pigeon update required, no change to this variable.  Pigeons can use any version equal to or greater than this number
2. The new version requires a Pigeon update
    * We update this variable to indicate the lowest pigeon version compatible with the new Paloma version.  This code goes through Paloma's upgrade proposal.  Pigeons can upgrade early if they want

ToasterTheBrave:
I'm marking this as ready for review.  After discussion in Discord, it sounds like we're aligned and agree on this strategy

app/app.go
taariq: Okay. so Anyone can fork paloma, update app.go, and submit their version for the new minimum pigeon. Yes?

taariq: I guess this is easier than having paloma VOTE on the minimum pigeon required, but it does remove governance for selecting the version.

taariq: Volume changing the number and then forcing a vote is kinda really volum controlling all the software. I think we need the version to be SET in governance from a community vote.

ToasterTheBrave: This minimum version does go through Paloma's voting though.  It's specified in the code, and the code is voted on during the upgrade proposal.  If the community doesn't like the change, they can vote against it just like any other code change in Paloma.

PR #925, opened 2023-07-20
Temporarily remove slashing functionality from Gravity

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/431

# Background

Until we finish adding the gravity features to Pigeon, we need to disable slashing in the gravity module.  We want this back, so it's just commented out for now to prevent buggy behavior in testnet.

# Testing completed

# Breaking changes


taariq (approved):
TEMPORARILY!

PR #936, opened 2023-07-31
refactor: Removing legacy code and restructuring

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/159

# Background

This change removes some of the legacy code around contract deployment that's not being used and very out of date by now. It also extracts all smart contract logic into a seperate file to add a bit seperation of concern.

Another addressed topic here is the differentiation between the compass EVM contract and arbitrary contracts deployed by validators without undergoing governance.

The change also includes scaffolding for unilateral contract deployment.

# Testing completed

# Breaking changes


taariq:
Yes, but this requires the use of Gas Management for the fee calculation and collection. This is the part that requires that the user pays for fee when using this. I am assuming that no fee is called for and just allows anyone to call this and then drain the validators. Let's discuss, but we may need to close until fee management integration. We really want to be able to charge fees for this.

taariq:
I changed my mind. Let's merge it and then upgrade. @ToasterTheBrave this feature needs gas management. let's talk abut what's required to integrate.

byte-bandit:
@taariq This is only the first part of the implementation. There is no contract actually being uploaded yet, no validator work being done. It should be safe to merge.

taariq:
Thanks @byte-bandit. Merged. Let's keep moving.

PR #940, opened 2023-08-08
Validators must support all active chains

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/608

# Background

Previously, we were only requiring that a validator be unjailed and bonded to be in the new valset.  Now, we are actually checking that they support all active chains.  This prevents validators that are artifically keeping themselves unjailed from getting into the active valset.

# Testing completed

# Breaking changes


byte-bandit (approved):
LGTM

PR #942, opened 2023-08-09
Publish valset to all chains when returning from a software update

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/608

# Background

Adding the ability to force publish a valset to all chains.  Then taking this ability and applying it to when we come back from an upgrade.

We are very often just above 2/3 previous consensus when we come back from an upgrade, so we should publish the valset that has changed by that much, in case any validators have trouble right after the upgrade and go offline.

Adding some test cases adjacent to this new code, just to test the different paths

There is also a small logging fmt bug fix in here because it finally annoyed me too much.

# Testing completed

# Breaking changes

### DIFF
--- a/app/upgrades.go
+++ b/app/upgrades.go
@@ -181,6 +181,23 @@ func (app *App) RegisterUpgradeHandlers(semverVersion string) {
 				return vm, err
 			}
 
+			// Try to build a snapshot.  Fails silently if unworthy
+			_, err = app.ValsetKeeper.TriggerSnapshotBuild(ctx)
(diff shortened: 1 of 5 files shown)


ToasterTheBrave:
Leaving this as draft for the moment.  I tested it in a local testnet yesterday, but I've had some small tweaks while writing tests today.  I need to do a full functional test tomorrow before this PR is ready to be merged.

byte-bandit (approved):
LGTM

PR #943, opened 2023-08-10
Cranking down a timeout to crank up chain speed

ToasterTheBrave:
# Related Github tickets

# Background

Overriding a config setting to set block timeout to 1s for all validators.  This will speed up block production by 4 seconds at the risk of fewer (but still at least 2/3) validators getting signatures in.

# Testing completed

# Breaking changes


byte-bandit (approved):
LGTM, I hope they can keep up.

PR #944, opened 2023-08-10
Add cli command to delete any stuck smart contract deployments

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/610

# Background

Paloma tracks smart contract deployments purely as a caching mechanism to prevent double-deploying too quickly.  We have an edge case where a deployment fails to go through for too long, it will get stuck.  Paloma will think it's deploying, but the message was long since deleted.

Now, we have a way of deleting this stuck deployment.  This is effectively the same as clearing the cache flag, and the chain will trigger a fresh deployment on the next block.

# Testing completed

# Breaking changes

### DIFF
--- a/proto/palomachain/paloma/evm/tx.proto
+++ b/proto/palomachain/paloma/evm/tx.proto
@@ -5,6 +5,7 @@ option go_package = "github.com/palomachain/paloma/x/evm/types";
 
 service Msg {
     rpc DeployNewSmartContract(MsgDeployNewSmartContractRequest) returns (DeployNewSmartContractResponse);
+    rpc RemoveSmartContractDeployment(MsgRemoveSmartContractDeploymentRequest) returns (RemoveSmartContractDeploymentResponse);
 }
 
 message MsgDeployNewSmartContractRequest {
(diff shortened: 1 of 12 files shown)


taariq (commented):
Delete and redeploy. However, this can be used by anyone? If so, is this an attack vector on contract deployment at the CLI? Someone might just keep hitting the CLI endpoint to watch the world burn?

byte-bandit:
What about reacting upon the message getting deleted? It's Paloma that removes it from the chain once it expires, right? Wouldn't we be able to update the deployment store accordingly? That way we don't have to introduce some manual override and can simply wait for the message to expire and a re-trigger should happen automatically. Sort of like the outbox pattern.

ToasterTheBrave:
@taariq No, I don't believe this can be exploited for an attack.  The deployment in this store is basically just used as a cache to check if we have a deployment in progress.  The deployment has already kicked off.  You can't prevent deploying by spamming this delete cli endpoint.  All you would potentially do is trigger a second deploy, which would have minimal effect because Pigeon actually wouldn't process the second deploy.

@byte-bandit Yes, we should be removing the deployment when deleting the message.  That's the better long term fix for this.  However, that's a good bit more code and is also a change, rather than new functionality.  I opted for this approach because we're going straight to mainnet with this release.  

We basically need something fast and unlikely to break so that we can hotfix it into mainnet.  That's what this is, and I think it's the right approach weighing the concerns of moving quickly vs not testing in testnet.

byte-bandit (approved):

PR #945, opened 2023-08-10
Increase the number of blocks to keep stale messages

ToasterTheBrave:
# Related Github tickets

# Background

We just increase our block speed drastically.  As such, we need to increase the amount of blocks to allow messages to remain stale. This is because we have more blocks per minute.

We also don't need to check quite as often for the same reason.

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/module.go
+++ b/x/consensus/module.go
@@ -164,8 +164,8 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.Val
 		am.keeper.Logger(ctx).Error("error while attesting to messages", "err", err)
 	}
 
-	if ctx.BlockHeight()%10 == 0 {
-		err := am.keeper.DeleteOldMessages(ctx, 300)
+	if ctx.BlockHeight()%50 == 0 {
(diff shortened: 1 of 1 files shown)


taariq (approved):

byte-bandit (approved):
I would also be interested in performance metrics in general. There's a lot of stuff happening on every block, especially I see a lot of unoptimised/unbatched kv store reads/writes, and we just upped execution speed by 500% ish.

taariq:
@byte-bandit any recommendations for monitoring and evaluating call performance?

byte-bandit:
@taariq Traditionally I'd use some tracing in each modules end block logic, with that telemetry being sent and processed on services like Datadog or Grafana. 

With a decentralised system, maybe a `monitoring` module is the more idiomatic approach. Something that'll hook into the end block logic and track time, storing the results with each block on the chain so they're retrievable by anyone. 

However, since anyone can run a node and we have little to no information what kind of hardware each node is running on, I don't think those results will be too reliable. Reproducible profiling as part of CI benchmarking would be better, but that's something for a full time devops role.

PR #947, opened 2023-08-14
feat: prevent pigeon liveness check for non-validators

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/643

# Background

This change will prevent a liveness check against Pigeon on clients which are not configured against a node on localhost in order to include non-validator CLI users. The liveness check will still be active for validators running CLI commands against their own hosts.

# Testing completed

# Breaking changes

### DIFF
--- a/cmd/palomad/root.go
+++ b/cmd/palomad/root.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/pprof"
 	"os"
+	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


taariq:
@ToasterTheBrave will you please review, approve, merge, and cut a new release that we will be able to ship to cli users?

ToasterTheBrave (approved):

PR #948, opened 2023-08-14
Revert 1s speed change

ToasterTheBrave:
# Related Github tickets

# Background

We are seeing a problem in testnet now that we have upgraded.  The most likely culprit we have identified is this increased block speed.  It seems we are moving too quickly for pigeons to initially rejoin after the chain has resumed.  They are jailed too quickly.

The fastest path to fixing this, as well as to preventing the problem on mainnet is to roll back the changes until we can test more thoroughly.

# Testing completed

# Breaking changes


byte-bandit (approved):

taariq:
so agreed this is NOT a speed bug?

ToasterTheBrave:
Agreed.  The speed does not appear to be what caused our issue in testnet.  Closing this

PR #962, opened 2023-08-29
feat: Migrate Pigeon TTL to block height

byte-bandit:
# Related Github tickets

https://github.com/VolumeFi/paloma/issues/707

# Background

The previous Pigeon keep alive implementation would use a five minute TTL and compare against the current block time. This would frequently lead to problems after a chain halt, during which Pigeons would not be able to refresh their TTL.

This change removes the time based approach and substitutes the block time with block height, with a default of 185 blocks of lifetime before a Pigeon is considered stale.

# Testing completed

# Breaking changes


x/valset/keeper/keep_alive.go
```diff
@@ -55,27 +57,38 @@ func (k Keeper) IsValidatorAlive(ctx sdk.Context, valAddr sdk.ValAddress) (bool,
...
 	if err != nil {
-		return time.Time{}, err
+		return 0, err
+	}
+
+	// hack hack hack
+	// To ensure a smooth migration to block based Pigeon TTL, we're going to pretend we received a valid
+	// future block height as long as the AliveUntil time is still valid.
+	// remove this once https://github.com/VolumeFi/paloma/issues/709 has been deployed to all nets
+	if data.AliveUntilBlockHeight == 0 {
```
ToasterTheBrave, line 81: Good catch

ToasterTheBrave (approved):
Looks great!

byte-bandit:
> On further review, I think we need to increment the required Pigeon version and have an associated Pigeon PR with this.
> 
> When Pigeon checks the keep alive, it's expecting a time back. https://github.com/palomachain/pigeon/blob/master/relayer/keep_alive.go#L13

That's true, but it still works as expected. Pigeons will still report in (due to the time stamp always being lower), it works without issues in test net. Although the frequency is elevated of course.

You're correct that we should handle this properly, but we'll need the Paloma update to go through first to reference the updated Proto definitions (unless we want to work against my feature branch, which I think is kind of messy), and if I increase the minimum Pigeon version now, we aim at a non-existing version. 

So I'd rather follow up with that in a separate PR. What do you think?

PR #963, opened 2023-08-31
feat: 🔧 prevent jailing high stake validators

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/708

# Background

The only security measurement in place during jailing so far was a check to see that Paloma would not run with less than 1 validator. 

This change adds a new safeguard, making sure that a validator will not be jailed if they're holding 25% or more of the entire network stake.

# Testing completed

# Breaking changes

### DIFF
--- a/x/valset/keeper/keeper.go
+++ b/x/valset/keeper/keeper.go
@@ -24,20 +24,21 @@ import (
 const (
 	snapshotIDKey                   = "snapshot-id"
 	maxNumOfAllowedExternalAccounts = 100
+	cJailingNetworkShareProtection  = 0.25
 )
 
 type Keeper struct {
-	cdc        codec.BinaryCodec
-	storeKey   storetypes.StoreKey
-	memKey     storetypes.StoreKey
-	paramstore paramtypes.Subspace
-	staking    types.StakingKeeper
-	ider       keeperutil.IDGenerator
-	EvmKeeper  types.EvmKeeper
-
+	EvmKeeper         types.EvmKeeper
 	SnapshotListeners []types.OnSnapshotBuiltListener
 
+	cdc                  codec.BinaryCodec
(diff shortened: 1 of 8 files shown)


x/valset/keeper/keeper.go
```diff
@@ -525,16 +528,26 @@ func (k Keeper) Jail(ctx sdk.Context, valAddr sdk.ValAddress, reason string) err
...
 			count++
 		}
 		return false
 	})
+
 	if count == 1 {
 		return ErrCannotJailValidator.Format(valAddr).WrapS("number of active validators would be zero then")
 	}
+
+	if float64(consensusPower)/float64(totalConsensusPower) > cJailingNetworkShareProtection {
```
ToasterTheBrave, line 547: Oh, it's done per-validator?  We could still hit a case where 2 validators are jailed at the same time to bring us below consensus.  Would it be more desirable to prevent the entire jailing event if we're about to jail validators summing 25% of consensus?

byte-bandit: `totalConsensusPower` is recalculated for every jail event. 

In case we have 5 validators, each with a stake of `20`, and three of them are suddenly without a Pigeon and should be jailed. What would happen is:

Validator `A` holds 20 stakes (out of 100). It's below 25%. Validator is jailed.
Validator `B` holds 20 stakes (out of 80). That's exactly 25%. Validator is jailed.
Validator `C` and the others holds 20 stakes (out of 60). That's above 25%. Validator is not jailed.

ToasterTheBrave: That makes total sense and should protect us then from jailing into a chain halt 👍

ToasterTheBrave (approved):

PR #964, opened 2023-08-31
Gravity Rewrite

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/431

# Background

A complete rewrite of Gravity to use the functionality from Gravity-Bridge instead of PeggyJV

This strips down the unneeded functionality from Gravity and only includes the parts we need.
I have also removed duplicate functionality for things like valset and validator eth address lookups.  We rely on our existing workflows for those.

The new functionality here includes:
* Vote through governance to track a token as the official token on each supported chain
* Submit "send-to-eth" transactions to send from a paloma account to an evm account
* Batching of transactions every 50 blocks
* Observer "send-to-paloma" events to add to a paloma account when deprecating from an evm account

I'm opening this as draft for now.  It's a huge amount of code.  The best way to test it is likely to stand it up in the private testnet and verify that everything works as expected.

# Testing completed

# Breaking changes


byte-bandit (changes requested):
Throwing in a blocker for now until we have `1.7.3`, I think this should go into `1.8.0`.

byte-bandit (approved):
This PR is simply too large to be reviewed sensibly, it's impossible to tell which part are custom adaptions and which is directly taken from the original gravity module.

To be honest, I think we'd be going back and forth for a week or two defining common coding guidelines for the Paloma project and bringing everything up to par. It's something I'd like to do given the time to reduce the clutter from merging in foreign code, but it will completely kill project velocity.

PR #966, opened 2023-09-01
Fix signing of gravity batches

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/517

# Background

This was put off because I didn't fully understand how Compass hashed batches before comparing signatures.  Now, I've come back to this and fixed it to sign exactly like Compass expects.

The test used here is using a real example, to ensure it works exactly as expected.

# Testing completed

# Breaking changes


taariq (approved):
LGTM!

PR #968, opened 2023-09-04
Add release notes

byte-bandit:


taariq (approved):
More validator happiness

Issue #973, opened 2023-09-12
Pigeon Jailed Report from multiple Validators


taariq:
# What is happening?
<details>
  <summary>Section description</summary>
<i>Provide as much context as you can. Give as much context as you can to make it easier for the developers to figure what is happening.</i>
</details>

Pigeon jail reports from validators are here:
from NodeJumper
```
time="2023-09-11T22:59:44+02:00" level=info msg="attesting 1 messages" action=attest message-ids="[96093]" messages-to-attest="[96093]" queue-name=evm/bnb-main/validators-balances x-correlation-id=cjvnvg0ab4l6ib3q7aqg
time="2023-09-11T22:59:44+02:00" level=debug msg="start processing validator balance request" queue-type-name=evm/bnb-main/validators-balances x-correlation-id=cjvnvg0ab4l6ib3q7aqg
time="2023-09-11T22:59:45+02:00" level=info msg="checking if validator is staking"
time="2023-09-11T22:59:45+02:00" level=info msg="event watcher loop" x-correlation-id=cjvnvg8ab4l6ib3q7asg
time="2023-09-11T22:59:45+02:00" level=info msg="signer loop"
time="2023-09-11T22:59:45+02:00" level=info msg="relayer loop" x-correlation-id=cjvnvg8ab4l6ib3q7at0
time="2023-09-11T22:59:45+02:00" level=info msg="event watcher loop" x-correlation-id=cjvnvg8ab4l6ib3q7au0
time="2023-09-11T22:59:45+02:00" level=info msg="validator is staking"
time="2023-09-11T22:59:46+02:00" level=debug msg="got height for time" height=31658960 nearest-to-time="2023-09-11 20:59:08.414592334 +0000 UTC" queue-type-name=evm/bnb-main/validators-balances x-correlation-id=cjvnvg0ab4l6ib3q7aqg
time="2023-09-11T22:59:46+02:00" level=info msg="got balance" balance=79445768000000000 evm-address=0xf9F9176a96017139ED75F91F047ec11574826304 height=31658960 nearest-to-time="2023-09-11 20:59:08.414592334 +0000 UTC" queue-type-name=evm/bnb-main/validators-balances x-correlation-id=cjvnvg0ab4l6ib3q7aqg
[...]

PR #979, opened 2023-09-15
feat: pigeon reponse data logging

byte-bandit:
# Related Github tickets

- https://github.com/palomachain/paloma/issues/973

# Background

We often run into situations where response data from handling messages is lost as said messages leave the queue, leaving us to guess what happened. This change adds the response data to the Paloma log to be queried at a later stage in order to enrich the available information we have at our disposal.

# Testing completed

# Breaking changes


taariq (approved):
LGTM!

PR #980, opened 2023-09-15
chore: additional logging in message attesting

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/778

# Background

When a new compass deployment fails, Paloma should automatically remove the deployment (and restart the process). The deployment however isn't removed, which leads to a stuck deployment. 

There's no indication why the deployment isn't removed. This together with the additional logging from https://github.com/palomachain/paloma/pull/979 should give us enough information to understand what's happening.

# Testing completed

# Breaking changes


taariq (approved):
Log it good.

taariq:
@byte-bandit checks still failing.

taariq (approved):

PR #981, opened 2023-09-18
Add pigeon status updates to Paloma

byte-bandit:
# Related Github tickets

# Background

There are some cases in which we're still lacking proper visibility, especially during operations running on individual Pigeons that do not return an error in order to enable retries, i.e. during message processing without consensus.

This change adds a new gRPC endpoint to Paloma that will do nothing else aside from adding the given string in a log message for us to intercept. The behaviour is guarded behind a feature flag in order to disable it by default and prevent malicious flooding of nonsense messages.

# Testing completed

# Breaking changes

### DIFF
--- a/proto/palomachain/paloma/paloma/tx.proto
+++ b/proto/palomachain/paloma/paloma/tx.proto
@@ -3,5 +3,22 @@ package palomachain.paloma.paloma;
 
 option go_package = "github.com/palomachain/paloma/x/paloma/types";
 
-// Msg defines the Msg service.
-service Msg {}
+service Msg {
+  rpc AddStatusUpdate(MsgAddStatusUpdate)
+    returns (EmptyResponse);
+}
+
+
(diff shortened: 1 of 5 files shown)


taariq (approved):
LFG.

ToasterTheBrave (commented):
Looks like a very sensible way of passing log messages into paloma's log.  Let's just make sure we test it in a testnet with some nodes logging and some not.  

This is the type of behavior that can cause app hash mismatch errors.  It doesn't look like this will because it's just writing logs, but let's test just to be sure.  Any time it's possible for validators to do things differently from each other, we risk those errors if anything's ever written.

taariq:
agreed @ToasterTheBrave.

byte-bandit:
@ToasterTheBrave @taariq I used a different branch `clohr/msg-meta-test` on which I built a CLI command to test this. It works as intended.

PR #985, opened 2023-09-20
Expose a query endpoint for latest published valset

ToasterTheBrave:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/783

# Background

Pigeon currently gets this information by querying the compass contract. The problem is, when there is a just-in-time valset update, pigeon will end up using an old valset because the new valset hasn't been committed yet.  We need the newest valset that may still be pending on the evm chain.

Now, we can query paloma instead of the evm contract.  Paloma knows what the latest published valset is, even if it hasn't yet been committed.

# Testing completed

# Breaking changes


x/valset/client/cli/query_get_latest_published_snapshot.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/palomachain/paloma/x/valset/types"
+	"github.com/spf13/cobra"
+)
+
+func CmdGetLatestPublishedSnapshot() *cobra.Command {
+	cmd := &cobra.Command{
```
byte-bandit, line 11: we dont need the cli command for pigeons to pick up this info, but it won't hurt either :)

byte-bandit (approved):
lgtm

x/valset/keeper/grpc_query_get_latest_published_snapshot.go
```diff
@@ -0,0 +1,26 @@
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	snapshot, err := k.GetLatestSnapshotOnChain(ctx, req.ChainReferenceID)
+	if err != nil {
+		return nil, err
+	}
+
+	return &types.QueryGetLatestPublishedSnapshotResponse{
+		Snapshot: snapshot,
+	}, nil
+}
```
byte-bandit, line 26: Nitpicking, low hanging fruit to improve test coverage here.

ToasterTheBrave: I'm working on the tests right now

x/valset/client/cli/query_get_latest_published_snapshot.go
ToasterTheBrave: Super easy to include the CLI command.  I didn't see any harm in it and it makes it easier to debug

x/valset/keeper/grpc_query_get_latest_published_snapshot.go
ToasterTheBrave: Tests added

PR #992, opened 2023-09-25
update libwasm

byte-bandit:
This brings back the libwasm wrapper update. Confirmed on test net that it works with the correct version of the library deployed.


byte-bandit:
will use dependabot instead

PR #996, opened 2023-09-28
feat: retry failed SLC calls with different Pigeons

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/802

# Background

This change will re-create a SubmitLogicCall action on the consensus queue in case it has not been retried yet. It will be retried once, Paloma will restart validator selection. The same pigeon might be selected again, but it's a start.

# Testing completed

# Breaking changes


byte-bandit:
Works like a charm in private test net.

verabehr (approved):

Issue #795, opened 2023-04-12
Cosmos-SDK V0.47.1 Upgrade: Bring Pigeon dependencies to Cosmos-SDK v0.47.1


taariq:
# Background
After successful upgrade of Paloma to Cosmos-SDK v0.47.1, we noticed that [pigeon ](https://github.com/palomachain/pigeon) start was failing with the error:
```
Apr 12 17:50:52 testnet-1 pigeon[1308114]: {"level":"info","msg":"Starting healthcheck server","time":"2023-04-12T17:50:52Z"}
Apr 12 17:50:52 testnet-1 pigeon[1308114]: panic: Bytes left over in UnmarshalBinaryLengthPrefixed, should read 10 more bytes but have 152
Apr 12 17:50:52 testnet-1 pigeon[1308114]: goroutine 1 [running]:
Apr 12 17:50:52 testnet-1 pigeon[1308114]: github.com/palomachain/pigeon/chain/paloma.getMainAddress({0xc000ccc0b0, {{{0xc000cb8c88, 0x4}, {0xc000cb8d28, 0x6}, {0xc000cb
/usr/local/bin/pigeon version
App version: v0.11.7
Build commit hash: 958ed17722fc2e468716276883a1cc4052ba07ca
```
For example: https://github.com/palomachain/pigeon/blob/master/chain/paloma/client.go#L25 still refers to tendermint and Paloma is now on CometBFT. Pigeon healthcheck appears to require tendermint, so we now need to upgrade Pigeon to Cosmos-SDK v0.47.1 as Paloma.

# Done when
- [ ] Update Pigeon to use Cosmos-SDK v0.47.1 packages: For example migrate to cometbft from tendermint.
- [ ] Test Pigeon https://github.com/palomachain/pigeon  is able to start after upgrade.

taariq:
Added links to https://github.com/palomachain/pigeon

0xmuralik:
Need to upgrade cosmos-sdk to v0.47.0 in https://github.com/VolumeFi/lens-1.
Refer to https://github.com/strangelove-ventures/lens/commit/735a2a0195eb0a9c0eb7a0f54813bf9f0b09b433.
This is a dependency in pigeon.

verabehr:
depends on https://github.com/VolumeFi/paloma/issues/115

taariq:
Fixed with https://github.com/palomachain/pigeon/pull/138. Closing.

PR #1002, opened 2023-10-04
feat!: #795 Transfer ERC20 ownership on compass update

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/795

# Background

    Since compass is the owner of any ERC20 tokens deployed to target
    chains, we need to make sure we transfer the ownership to the
    new compass during the upgrade process.
    
    This change alters the current upgrade procedure. Once the new
    compass is reported as deployed, we send a new message to Pigeon,
    instructing it to perform the transfer. Only after we get a
    successful ownership transfer attestation will we change
    Paloma to use the new compass contract.
    
    Also contains a small wrapper for the default logging framework,
    as I found no default reliable way of logging errors as well as
    some housekeeping.

# Testing completed

# Breaking changes


taariq (approved):
Looks very logical and a sane process. LGTM and test!

PR #1010, opened 2023-10-10
fix: ease keep-alive duration

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/856

# Background

Increasing the TTL to approx. 50 minutes. Pigeons will start calling in 15 minutes before. This should give enough leeway to allow us to bring back locking and avoid the `account sequence mismatch` error.

# Testing completed

# Breaking changes


taariq (approved):
LGTM

PR #1011, opened 2023-10-11
JIT Valset update for all jobs

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/846

# Background

  Previously, we published a JIT valset update only for jobs
  that were triggered using the scheduling module. Bots
  and other contract triggering job authors would simply
  bypass this.

  This change adds a JIT valset update to any job, no matter
  the triggering call path.

  It also includes an improvement to return with the ID of the
  consensus message that was created upon executing a job.
  This information will be stored as part of the TX, giving us a
  full link into which TX lead to which message being triggered.

# Testing completed

# Breaking changes


byte-bandit:
Drafting this to block merge until I resolve contract deployment issues due to the new erc20 ownership transfer.

verabehr (approved):

PR #1015, opened 2023-10-17
feat: hold SLC until JIT valset update is processed

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/855

# Background

This change adds an additional filter when Pigeons ask for any messages for relaying. It will hold off on returning any SLC messages on target chains that have pending valset updates.

This takes care of two race conditions:

- ValsetUpdate & SLC deployed by same Pigeon: replaying of both messages happens VERY quickly in succession -> high risk of `Incorrect Checkpoint` error
- ValsetUpdate & SLC deployed by two different Pigeons: even higher risk, as the SLC might get relayed BEFORE the valset update

With this change, SLC messages will never be given out to Pigeons as long as a previous valset update is still outstanding, meaning there is no way it would get delivered before the valset update or in too close a succession.

# Testing completed

# Breaking changes


taariq (approved):
LGTM

PR #1023, opened 2023-11-02
fix: assigning msgs to validators w/o chain support

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/944

# Background

There's a hen & egg situation in case we're currently onboarding a new chain on Paloma, as the chain status is not yet active, Validators without full support for the new chain will still be within the active validator set.

This change adds target chain awareness when filtering for viable message relayers, ensuring that validators with missing chain support for the required target chain are no longer considered.

# Testing completed

# Breaking changes


taariq (approved):
A small fix. A big impact.

PR #1025, opened 2023-11-06
feat: introduce basic multisig message support

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/857
- https://github.com/VolumeFi/paloma/issues/934

# Background

This change introduces a couple of new concepts into the Paloma ecosystem:

1. Message Metadata: every message that is to be included within a transaction now comes with a new, required field called `metadata`, which itself includes two different fields: `signers` and `creator`.
2. Ante Handlers: Paloma's modules may now register their own ante handlers to perform transaction validation. 

Using these new concepts, Paloma now considers the metadata field contents when verifying transaction signatures. It does so by confirming that the transaction contains signatures from:

a) all signers of every message within the transaction
b) ensuring that the message creator is among the set of signers, or **at least** one of the signers of a message has active feegrant from the message creator

This allows us to start using more than one key when signing transactions, as long as those keys are feegranted from our validator account. It also sets the stage for multisig transaction verifications on our end.

⚠️ This is a breaking change, and will require a new version of Pigeon to be deployed alongside.

# Testing completed

# Breaking changes


x/paloma/ante.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		logger(ctx).Debug(fmt.Sprintf("got %d allowances from granter %s", len(grants.GetAllowances()), creator))
+		grantsLkUp := map[string]feegrant.Grant{}
+		for _, v := range grants.GetAllowances() {
+			if v == nil {
+				continue
+			}
+
+			logger(ctx).Debug("found allowanec", "granter", v.GetGranter(), "grantee", v.GetGrantee())
```
byte-bandit, line 122:
```suggestion
			logger(ctx).Debug("found allowance", "granter", v.GetGranter(), "grantee", v.GetGrantee())
```

verabehr (approved):

byte-bandit:
> LGTM! Do we need README updates as well?

Not for this one, the Pigeon one will contain one though.

PR #1027, opened 2023-11-06
fix: gravity message logging format

byte-bandit:
The current logging format is pretty useless and doesn't allow to query for transactions via event predicates. This should help.


verabehr (approved):

taariq:
@byte-bandit thanks for this update.

PR #1029, opened 2023-11-10
feat: add query for last observed event nonce

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/959

# Background

This adds a new GRPC query to retrieve the last observed event nonce from Paloma. Pigeons will use this information in order to build their own state.

# Testing completed

- Will be tested in TN together with the Pigeon change

# Breaking changes


taariq (approved):
This seemed way more SIMPLER than the discussion around this issue. Is there more required? LGTM

byte-bandit:
> This seemed way more SIMPLER than the discussion around this issue. Is there more required? LGTM

Complex problem, simple solution ;)

Pigeon will need an update to consume this information, but it won't be very complex.

PR #1035, opened 2023-11-22
feat: update paloma app to SDK v0.50

deepan95dev:
# Background

Upgraded the files to 0.50 other than x

# Testing completed

- [x] No Import errors
- [x] No Usage errors


app/export.go
```diff
@@ -16,7 +16,7 @@ import (
 // file.
 func (app *App) ExportAppStateAndValidators(forZeroHeight bool, jailAllowedAddrs, modulesToExport []string) (servertypes.ExportedApp, error) {
 	// as if they could withdraw from the start of the next block
-	ctx := app.NewContext(true, tmproto.Header{Height: app.LastBlockHeight()})
+	ctx := app.NewContext(true /*tmproto.Header{Height: app.LastBlockHeight()}*/)
```
aleem1314, line 19: Why second argument commented?

cmd/palomad/root.go
```diff
@@ -185,28 +192,83 @@ func rootPreRunE(cmd *cobra.Command, args []string) error {
...
+}
+
+// newApp creates the application
+func newApp(
+	logger log.Logger,
+	db dbm.DB,
+	traceStore io.Writer,
+	appOpts servertypes.AppOptions,
+) servertypes.Application {
+	//baseappOptions := server.DefaultBaseappOptions(appOpts)
```
aleem1314, line 243: ditto here

app/app.go
```diff
@@ -8,31 +8,45 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/CosmWasm/wasmd/x/wasm"
-	wasmkeeper "github.com/CosmWasm/wasmd/x/wasm/keeper"
-	dbm "github.com/cometbft/cometbft-db"
+	// storetypes "cosmossdk.io/store/types"
```
aleem1314, line 11:
```suggestion
```

app/app.go
```diff
@@ -41,9 +55,13 @@ import (
 	"github.com/cosmos/cosmos-sdk/x/bank"
 	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
-	"github.com/cosmos/cosmos-sdk/x/capability"
-	capabilitykeeper "github.com/cosmos/cosmos-sdk/x/capability/keeper"
-	capabilitytypes "github.com/cosmos/cosmos-sdk/x/capability/types"
+
+	"github.com/cosmos/ibc-go/modules/capability"
+	// ica "github.com/cosmos/ibc-go/v3/modules/apps/27-interchain-accounts"
```
aleem1314, line 60: ditto here

cmd/palomad/root.go
```diff
@@ -93,7 +100,7 @@ func NewRootCmd() *cobra.Command {
 		flags.FlagKeyringBackend: keyring.BackendOS,
 	}); err != nil {
 		// seems to be non-fatal issue
-		log.Printf("error overriding flag defaults: %s", err.Error())
+		fmt.Println("error overriding flag defaults: %s", err.Error())
```
byte-bandit, line 103: Why not use the log package?

app/export.go
vishal-kanna: NewContext takes only one argument which is of type bool.

cmd/palomad/root.go
deepan95dev: We imported `"cosmossdk.io/log"`  which does not have Printf function so its updated to fmt.

deepan95dev: pushed the changes with "log" module

aleem1314:
Looks good to me. Let's update each Paloma module in its own pull request and then integrate them with the Paloma app.

cmd/palomad/root.go
byte-bandit: We don't encourage using the `fmt()` package for logging. You can use `log.Info()` instead.

PR #1043, opened 2023-11-24
feat : update liblog to use SDKv0.50.0

ruthishvitwit:


x/paloma/ante.go
```diff
@@ -78,7 +78,7 @@ func NewVerifyAuthorisedSignatureDecorator(fk types.FeegrantKeeper) VerifyAuthor
 // AnteHandle verifies that the message is signed by at least one signature that has
 // active fee grant from the creator address, IF the message contains metadata.
 func (d VerifyAuthorisedSignatureDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
-	if simulate || ctx.IsCheckTx() {
```
aleem1314, line 81: Why this change?

taariq: @aleem1314 https://github.com/palomachain/paloma/pull/1040

ruthishvitwit: @aleem1314 sorry as i made a mistake by raising a pr to master branch and then changing it to v.50 upgrade branch this commit got pulled into this pr.

aleem1314 (approved):

util/liblog/liblog.go
```diff
@@ -13,6 +13,9 @@ type Logr interface {
 func FromSDKLogger(l log.Logger) Logr {
 	return &lgwr{l: l}
 }
+func (l *lgwr) Impl() interface{} {
+	return l
+}
```
byte-bandit, line 18: @aleem1314 Where does this come from? I think we can remove this.

aleem1314: In v0.50, `Impl()` method added to the `Logger` interface. So implementation required for each type. [Ref](https://github.com/cosmos/cosmos-sdk/blob/main/log/logger.go#L57)

byte-bandit: Well this is ugly AF ... Consider resolved.

PR #1047, opened 2023-11-28
feat: support event nonce ctor input for deployments

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1002

# Background

Adds support for setting the last observed event nonce from compass from within the constructor input during smart compass uploading. This allows updating the deployed compass contract without losing the current nonce value.

# Testing completed

# Breaking changes


byte-bandit:
This change does not require a new version of compass to be deployed. However, ONCE we re-deploy compass in the future, it will have to have support for this built in.

taariq (approved):
LGTM.

taariq:
> This change does not require a new version of compass to be deployed. However, ONCE we re-deploy compass in the future, it will have to have support for this built in.

@webelf101 do we need to make a PR for a new compass with this support so we can vote and redeploy compass next week?

byte-bandit:
@taariq No need to redeploy compass next week just for the sake of supporting this change. Should be all good. Though next time we do upgrade compass for whatever reason, it will need to support this.

verabehr:
I'm happy to open the ticket. I'm a bit confused what the actual change needed is though as @wc117 provided the correct ABI and said that no code change was needed for it. If this ABI is what is needed then what change is still pending?

PR #1052, opened 2023-12-05
feat: log inbox queries

byte-bandit:
# Related Github tickets

# Background

This change adds logging debug information for inbox queries.

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/keeper/query_queued_messages_for_relaying.go
+++ b/x/consensus/keeper/query_queued_messages_for_relaying.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/palomachain/paloma/util/liblog"
 	"github.com/palomachain/paloma/x/consensus/types"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


verabehr (approved):

byte-bandit:
The problem with this approach is that queries are not propagated across the network and only run against your local node. That means that we're only seeing the requests from our Pigeon, which doesn't exactly solve the problem at hand. 

The first thing I thought about was emitting an event on the chain instead. But that will lead to consensus failure, as not all nodes emit the event (only the queried one), so useless.

I don't think there's a way to capture this other than creating a transaction - which will increase our TX bandwith a lot (`number of validators` transactions per block, only for observability). Doesn't sound like a good solution either.

I will leave this as draft while I work on the remainder of https://github.com/VolumeFi/paloma/issues/1036 for now. But the reason we use queries for this is because they're fast and don't rely on the rest of the network. At the moment, we call twice per second. 

There's a different idea I have though, we can buffer the data in Pigeon and just send it out once per minute as log update. That should give us the insights that we want, just not in real time. I will look into it.

taariq:
agreed and aligned with non real time data

byte-bandit:
Done with https://github.com/palomachain/pigeon/pull/344

PR #1055, opened 2023-12-07
feat: reassign orphaned messages

byte-bandit:
# Related Github tickets

# Background

Every 10 blocks, Paloma will inspect all supported consensus queues
for messages that are older than 10 blocks and still not processed.
Any message that's found and meant to be relayed will be reassigned.

Because validators are picked random-deterministically, it's possible
(though unlikely) that the same validator will be picked again.

Future plans for Pigeon feed will prevent a double assignment.

# Testing completed

# Breaking changes


taariq (approved):
LGTM!

PR #1056, opened 2023-12-07
chore: reduce consensus msg lifetime

byte-bandit:
# Related Github tickets

# Background

Changes the lifetime of a consensus message to 300 blocks, approximately 8 minutes with an average block speed of 1.6 seconds per block.

Before automatic reassignment of messages, a message would stay in the queue for up to 1200 blocks, giving the assigned validator some time to fix ongoing issues.

With the reassignment of tardy messages in place, there's no need for messages to stay around considerably longer than 100 blocks really, but with 300 blocks we still get some opportunity to inspect what's going on in logs.

We can reduce this value further to something like 100 or even 50 once we're confident with our relay stability.

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/module.go
+++ b/x/consensus/module.go
@@ -171,7 +171,7 @@ func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.Val
 	}
 
 	if ctx.BlockHeight()%50 == 0 {
-		err := am.keeper.PruneOldMessages(ctx, 1200)
+		err := am.keeper.PruneOldMessages(ctx, 300)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


taariq:
Why not 5 minutes? does it matter?

taariq (approved):
LGTM!

byte-bandit:
@taariq 
> Why not 5 minutes? does it matter?

The longer message stay in the queue, the easier it is for us to debug ongoing issues in production. Once they're gone, it becomes more difficult. In theory though, we should have a lot of intel in the logs by now. So once I see this working as expected, I believe we can lower it considerably.

PR #1038, opened 2023-11-23
feat : upgraded the app with consensus module to SDK v0.50.0

ruthishvitwit:

PR #1060, opened 2023-12-15
feat: upgraded app with paloma module to SDK v0.50

vishal-kanna:


x/paloma/keeper/keeper_integration_test.go
```diff
@@ -108,18 +109,21 @@ var _ = Describe("jailing validators with missing external chain infos", func()
 
 		It("jails val[1] and val[2], but no val[0]", func() {
 			By("validators are not jailed")
-			Expect(a.ValsetKeeper.IsJailed(ctx, vals[0].GetOperator())).To(BeFalse())
-			Expect(a.ValsetKeeper.IsJailed(ctx, vals[1].GetOperator())).To(BeFalse())
-			Expect(a.ValsetKeeper.IsJailed(ctx, vals[2].GetOperator())).To(BeFalse())
+			valAddress1, err := keeperutil.ValAddressFromBech32(a.PalomaKeeper.AddressCodec, vals[0].GetOperator())
+			valAddress2, err := keeperutil.ValAddressFromBech32(a.PalomaKeeper.AddressCodec, vals[1].GetOperator())
+			valAddress3, err := keeperutil.ValAddressFromBech32(a.PalomaKeeper.AddressCodec, vals[2].GetOperator())
```
byte-bandit, line 114: So, this back and forth of parsing formats is getting out of hand, it even transpires to test cases.

We need a better solution for this. I'm open to ideas and suggestions here. I have to ideaas that come to mind:

### a) create util functions
that are used for getting operator/validator/consesus addresses, that assume getting the  keys will usually work and look somewhat like this:

```go
func (a *Addresser) GetOperator(v sdk.ValidatorI) string {
 defer func() { if r := recover(); r != nil {
  // log error & return default value
} }()

  return address and panic if error
}
```

It's not super pretty and basically copies the approach that was now removed in Cosmos of panicking all the time.

### b) Use proper types on the functions instead of `[]byte` or `string`

Might also be worth taking a look at other chains that have upgraded to `0.50` and check how they're avoid this DRY hell.

x/paloma/ante.go
```diff
@@ -112,21 +120,21 @@ func (d VerifyAuthorisedSignatureDecorator) AnteHandle(ctx sdk.Context, tx sdk.T
...
-		logger(ctx).Debug(fmt.Sprintf("got %d allowances from granter %s", len(grants.GetAllowances()), creator))
+		liblog.FromSDKLogger(logger(ctx)).Debug(fmt.Sprintf("got %d allowances from granter %s", len(grants.GetAllowances()), creator))
 		grantsLkUp := map[string]feegrant.Grant{}
 		for _, v := range grants.GetAllowances() {
 			if v == nil {
 				continue
 			}
 
-			logger(ctx).Debug("found allowance", "granter", v.GetGranter(), "grantee", v.GetGrantee())
+			liblog.FromSDKLogger(logger(ctx)).WithFields("granter", v.GetGranter(), "grantee", v.GetGrantee()).Debug("found allowance")
```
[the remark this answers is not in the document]
deepan95dev: Replaced the usage and pushed

x/paloma/keeper/keeper_integration_test.go
deepan95dev: I applied the same util function in [keeper](https://github.com/palomachain/paloma/pull/1060/commits/729f83611bec75d58cc4bdb0770af60772e9b848#diff-19d03101c764af0a94afb12fb3fa132115792735da677d8bae5b51d001ca2f39R87)  for the above case and pushed.

x/paloma/ante.go
```diff
@@ -78,30 +85,31 @@ func NewVerifyAuthorisedSignatureDecorator(fk types.FeegrantKeeper) VerifyAuthor
...
 		m, ok := msg.(libmeta.MsgWithMetadata[vtypes.MsgMetadata])
 		if !ok {
-			logger(ctx).Debug(fmt.Sprintf("msg %s does not contain metadata. skipping ownership verification...", proto.MessageName(msg)))
+			liblog.FromSDKLogger(logger(ctx)).Debug(fmt.Sprintf("msg %s does not contain metadata. skipping ownership verification...", proto.MessageName(msg)))
 			continue
 		}
 
 		creator := m.GetMetadata().GetCreator()
-		signers := msg.GetSigners()
+		signers := m.GetMetadata().GetSigners()
```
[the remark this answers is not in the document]
deepan95dev: `GetSigners()` was previously defined under `Metadata` now I defined under `MsgWithMetadata` too as suggested above and pushed.

x/paloma/keeper/keeper_integration_test.go
deepan95dev: I have gone through couple of chains' OSS but none of them have migrated to 0.50 yet.

byte-bandit: Did you look at all of them? Some of them might have dedicated feature branches like we do.

deepan95dev: I did not look at each, let me traverse through all of them.

deepan95dev: I came across Kujira-core chain which has upgraded SDK. They have just three occurences of `StringToBytes` conversion and did not create any util file for this case. Instead they created function `ValidatorAddressCodec` within keeper.go 

```go
// ValidatorAddressCodec returns the app validator address codec.
func (k Keeper) ValidatorAddressCodec() addresscodec.Codec {
	return k.validatorAddressCodec
}
```

byte-bandit: So, since CosmosSDK seems to have moved away from `string` representation to `[]byte`, I suppose it makes the most sense to update consuming function signatures to use `[]byte` as well instead of `string`.

Maybe worth creating custom types like `type ValAddr []byte` which allows us to implement stringer interface on it as well, also makes the intention of the `[]byte` variable a lot more explicit.

deepan95dev: I guess It's not moved away, we have `bech32` `string` type in client side tools like kepler wallet  /  chain explorers, whereas `byte` type is required for core sdk backend layer.

byte-bandit: Looks like lots of Paloma code is working against string representation. If Cosmos has moved to `[]byte`, we should follow. Obviously we can still support `string` on the frontend.

deepan95dev: I pushed the modifications as you mentioned in the top of this thread.

x/paloma/keeper/keeper.go
```diff
@@ -62,27 +63,46 @@ func NewKeeper(
 	}
 }
 
-func (k Keeper) Logger(ctx sdk.Context) log.Logger {
-	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
+func (k *Keeper) MustGetValAddr(addr string) (sdk.ValAddress, error) {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Printf("error while getting valAddr ", r)
```
aleem1314, line 69:
```suggestion
			log.Printf("error while decoding validator address ", r)
```

Use liblog here

aleem1314 (dismissed):

taariq:
@byte-bandit your approval required here.

util/libmeta/get_signers.go
```diff
@@ -11,6 +11,7 @@ type Metadata interface {
 
 type MsgWithMetadata[T Metadata] interface {
 	GetMetadata() T
+	GetSigners() []string
```
byte-bandit, line 14: Remove this. Use GetMetadata().GetSigners()

x/paloma/keeper/keeper.go
```diff
@@ -152,3 +155,15 @@ func (k Keeper) CheckChainVersion(ctx sdk.Context) {
 		return
 	}
 }
+func (k *Keeper) MustGetValAddr(addr string) sdk.ValAddress {
+	defer func() {
+		if r := recover(); r != nil {
+			k.Logger(context.Background()).Error("error while getting valAddr")
```
byte-bandit, line 161: Log the actual error

byte-bandit (changes requested):

util/libmeta/get_signers.go
deepan95dev: Reverted back to the old [commit](https://github.com/palomachain/paloma/pull/1060#discussion_r1431172964) `GetMetadata()`,

x/paloma/keeper/keeper.go
deepan95dev: done

x/paloma/keeper/keeper.go
```diff
@@ -152,3 +155,15 @@ func (k Keeper) CheckChainVersion(ctx sdk.Context) {
...
+		if r := recover(); r != nil {
+			k.Logger(context.Background()).Error("error while getting valAddr", r)
+		}
+	}()
+	bz, err := k.AddressCodec.StringToBytes(addr)
+	if err != nil {
+		panic(err)
+	}
+	return bz
+}
```
byte-bandit, line 169: You don't need to panic and recover within the same function. You can just react on the error. This really should live in a util package so other modules may use it as well.

Inject codec and address, return ValAddress or empty. Probably rename to `ValAddrOrEmpty` to make intention of this function clear.

x/paloma/keeper/keeper_integration_test.go
```diff
@@ -56,7 +53,8 @@ var _ = Describe("jailing validators with missing external chain infos", func()
 			// val[1] has only one chain
 			// val[2] doesn't have anything
 			// All other vals have everything
-			err := a.ValsetKeeper.AddExternalChainInfo(ctx, vals[0].GetOperator(), []*valsettypes.ExternalChainInfo{
+			valAddress, err := k.MustGetValAddr(vals[0].GetOperator())
```
byte-bandit, line 56: Does this even compile? `MustGetValAddr` doesn't return two types. It's also intended as a drop in replacement for `vals[0].GetValidator()` to reduce amount of boilerplate.

byte-bandit (changes requested):

util/libmeta/get_signers.go
deepan95dev: `msg.GetSigners()` used to exist in sdk.Tx `GetMsgs()` in earlier version, but after upgrade its moved, so we updated accordingly, later you have asked to use `m.getSigners` [here](https://github.com/palomachain/paloma/pull/1060#discussion_r1431172964) again we are back to use `msg.GetSigners`  [here](https://github.com/palomachain/paloma/pull/1060#discussion_r1437752513). Does it compile with `msg.GetSigners` as such method does not exist in SDK anymore?

x/paloma/keeper/keeper.go
deepan95dev: Injecting codec and address to return valAddress with error was [added here](https://github.com/palomachain/paloma/pull/1060/commits/44f60931a2dae4143e5274f9abd8b39f88caf8b0#diff-a62f6e247365c573943d95bdb8bd694c546f42d4431e2372c59d542b279775bdR5) and later removed as you suggested.  Should we revert that util function or any modification in it required?

x/paloma/keeper/keeper_integration_test.go
deepan95dev: no, there is not error returned from `MustGetValAddr`. Will fix and update this once we conclude on that util function as I mentioned [above](https://github.com/palomachain/paloma/pull/1060#discussion_r1437783431)

x/paloma/keeper/keeper.go
byte-bandit: I'd like to see a more comfortable way of working with this concept. The original util function you wrote was going in the right direction. Maybe we misunderstood each other.

The issue I has with it was that in [this state](https://github.com/palomachain/paloma/pull/1060/commits/44f60931a2dae4143e5274f9abd8b39f88caf8b0#diff-a62f6e247365c573943d95bdb8bd694c546f42d4431e2372c59d542b279775bdR5), it was nothing else but a wrapper - there was 0 value added. 

We agreed that an error was very unlikely to be returned, and I suggested to panic in case of an error. That whole logic was supposed to live in a util function, like this:

```go
myAddr := util.MustGetValAddr(string) // <-- will panic if the addr can't be returned
```

This is basically the way things worked with Cosmos `v0.47`, and I think it's more convenient that dealing with error handling every time.

That being said, I think I changed my mind on this. We should avoid a `panic` when we can, no matter the likelihood of occurrence, as it may result in a chain halt. That is most likely why Cosmos has removed `panic`s from their core code with `0.50` and relies on more idiomatic error handling now.

So, let's roll with them to stay closer to reference implementations and avoid `panic`. 

What does this mean for our situation? I still suggest to use a util function, that at least will ensure the address will get parsed the same way every time, and we can have an identical error message for it across modules, so it should be module and keeper agnostic, kind of like this:

```go
const ErrFailedToParseValAddress err = "failed to parse validator address from bech32"

type err string

func (e err) Error() string {
    return string(e)
}

func ValAddressFromBech32(addressCodec address.Codec, valAddr string) ([]byte, error) {
	bz, err := addressCodec.StringToBytes(valAddr)
        if err != nil {
            return nil, errors.Join(ErrFailedToParseValAddress, err)
        }
	return bz, nil
}
```

util/libmeta/get_signers.go
byte-bandit: We've had an extensive discussion about this on another PR before I believe.

You're correct that `GetSigners()` and `ValidateBasic()` are interfaces that were to be implemented by every message in `v0.47`, and this necessity has been removed with `0.50`, which is why you thought to remove it.

The issue is that Paloma has built a custom ante handler that relies on this interface. The ante handler will skip any messages that do not implement it, which is fine for all message types coming from Cosmos directly.

However, custom messages defined by Paloma still need to implement this interface so the ante handler picks them up accordingly.

This means:

- all custom messages need to explicitly implement `GetSigners()` and `ValidateBasic()`
- every implementation should work identically, using `libmeta`, and not access any other fields directly

There's more background in the comments on this PR: https://github.com/palomachain/paloma/pull/1038

x/paloma/ante.go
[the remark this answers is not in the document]
deepan95dev: Understood your point, here the custom message is not `msg` because it is cosmos-sdk's msg which does not have `GetSigners()` and the custom msg here  is `m.GetMetadata()` and that has `m.GetMetadata().GetSigners()`. Code is upto date with this usage, is anything else need to be added in this PR?  @byte-bandit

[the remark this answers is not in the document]
byte-bandit: That's ..... not good. Our entire setup for multiple signing keys is built on top of that interface, and the absence of it would break the ante handler.

Luckily, we don't care about common SDK messages and already skip the handling if the message fails to parse [here](https://github.com/palomachain/paloma/blob/2226e95be60c4972dff44afbbbe7b4cb0ec268c8/x/paloma/ante.go#L87)

Simply using `m.GetMetadata().GetSigners()` won't cut it here though. Use `libmeta.GetSigners(m)`, which performs some additional logic on top of simply reading the signer data.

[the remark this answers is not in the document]
deepan95dev: fine, pushed with `libmeta.GetSigners(m)`

x/paloma/ante_test.go
```diff
@@ -1,257 +1,258 @@
 package paloma_test
```
byte-bandit, line 1: Why is this test removed?

x/paloma/keeper/keeper.go
```diff
@@ -62,27 +63,32 @@ func NewKeeper(
...
 			mmap[mapkey{string(chainType), string(cri)}] = struct{}{}
 		}
 	}
 
 	var g whoops.Group
 	for _, val := range vals {
-		exts, err := k.Valset.GetValidatorChainInfos(ctx, val.GetOperator())
+		valAddr, err := utilkeeper.ValAddressFromBech32(k.AddressCodec, val.GetOperator())
+		if err != nil {
+			k.Logger(ctx).Error("Error while getting ValAddress", err)
```
byte-bandit, line 89: So we log the error, then what?

x/paloma/keeper/keeper.go
```diff
@@ -103,8 +109,12 @@ func (k Keeper) JailValidatorsWithMissingExternalChainInfos(ctx sdk.Context) err
 
 		sort.Strings(notSupported)
 
+		valAddr, err = utilkeeper.ValAddressFromBech32(k.AddressCodec, val.GetOperator())
+		if err != nil {
```
byte-bandit, line 113: Same here.

x/paloma/keeper/keeper_integration_test.go
```diff
@@ -56,7 +54,11 @@ var _ = Describe("jailing validators with missing external chain infos", func()
 			// val[1] has only one chain
 			// val[2] doesn't have anything
 			// All other vals have everything
-			err := a.ValsetKeeper.AddExternalChainInfo(ctx, vals[0].GetOperator(), []*valsettypes.ExternalChainInfo{
+			valAddress, err := utilkeeper.ValAddressFromBech32(k.AddressCodec, vals[0].GetOperator())
+			if err != nil {
```
byte-bandit, line 58: Okay, it's obvious you just copy/pasted this now. We don't need logging in test files. Just assert the error is `nil` and move on.

x/paloma/ante_test.go
deepan95dev: reverted this.

x/paloma/keeper/keeper.go
deepan95dev: I applied the same error catch as added [here](https://github.com/palomachain/paloma/blob/2226e95be60c4972dff44afbbbe7b4cb0ec268c8/x/paloma/keeper/keeper.go#L87)

x/paloma/keeper/keeper.go
deepan95dev: same catch here using whoops group [here](https://github.com/palomachain/paloma/blob/2226e95be60c4972dff44afbbbe7b4cb0ec268c8/x/paloma/keeper/keeper.go#L87)

x/paloma/keeper/keeper_integration_test.go
deepan95dev: Added assertion for error to be Nil

util/keeper/address.go
```diff
@@ -1,8 +1,22 @@
...
+)
+
+type err string
+
+const ErrFailedToParseValAddress err = "failed to parse validator address from bech32"
+
+func (e err) Error() string {
+	return string(e)
+}
 func ValAddressFromBech32(addressCodec address.Codec, valAddr string) ([]byte, error) {
```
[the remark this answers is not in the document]
deepan95dev: added address_test.go

util/keeper/address_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/codec/address"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/palomachain/paloma/testutil"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestValAddressConversion(t *testing.T) {
```
byte-bandit, line 11: - Verify return value matches expected value
- Add a test case for invalid address

x/paloma/keeper/keeper.go
```diff
@@ -103,8 +110,13 @@ func (k Keeper) JailValidatorsWithMissingExternalChainInfos(ctx sdk.Context) err
 
 		sort.Strings(notSupported)
 
+		valAddr, err = utilkeeper.ValAddressFromBech32(k.AddressCodec, val.GetOperator())
```
byte-bandit, line 113: Why are you calling it here again? `valAddr` is already loaded at this point, no?

byte-bandit (changes requested):
Mostly good now I think, two small things.

x/paloma/keeper/keeper.go
deepan95dev: yeah my bad, it might got added while resolving merge conflict. removed and pushed

util/keeper/address_test.go
deepan95dev: added these two cases with assertion

x/paloma/keeper/keeper.go
```diff
@@ -103,8 +110,12 @@ func (k Keeper) JailValidatorsWithMissingExternalChainInfos(ctx sdk.Context) err
 
 		sort.Strings(notSupported)
 
+		if err != nil {
+			g.Add(err)
+			continue
+		}
```
byte-bandit, line 116: This is unneeded, no?

byte-bandit (changes requested):

x/paloma/keeper/keeper.go
deepan95dev: should we ignore this err from `utilkeeper.ValAddressFromBech32` ? or any other handler needed?

byte-bandit: You define `err` in line 92: `exts, err := k.Valset.GetValidatorChainInfos(ctx, valAddr)`, and you perform the proper error checking on line 93.

As far as I can tell, `err` is not redefined before this code is reached, so this check is superfluous.

deepan95dev: so ignoring the `err` from valAddr is fine?  in line 87
````go
valAddr, _ := utilkeeper.ValAddressFromBech32(k.AddressCodec, val.GetOperator())
`````

Comment (byte-bandit) [reply x/paloma/keeper/keeper.go:116]: No. I'm not asking you to ignore any errors. Let me add some context to what I mean:

```go
for _, val := range vals {
                // err declared
		valAddr, err := utilkeeper.ValAddressFromBech32(k.AddressCodec, val.GetOperator())

                // err checked
		if err != nil {
			g.Add(err)
			continue
		}

                // err declared
		exts, err := k.Valset.GetValidatorChainInfos(ctx, valAddr)

                // err checked
		if err != nil {
			g.Add(err)
			continue
		}

                // err is not redeclared within the following code block
		valmap := make(map[mapkey]struct{})
		for _, ext := range exts {
			key := mapkey{ext.GetChainType(), ext.GetChainReferenceID()}
			valmap[key] = struct{}{}
		}

		notSupported := []string{}
		for mustExistKey := range mmap {
			if _, ok := valmap[mustExistKey]; !ok {
				// well well well
				notSupported = append(notSupported, fmt.Sprintf("[%s, %s]", mustExistKey[0], mustExistKey[1]))
			}
		}

		sort.Strings(notSupported)

                // superfluous error check. We already checked this value, and it hasn't been redeclared since.
		if err != nil {
			g.Add(err)
			continue
		}

		if len(notSupported) > 0 {
			g.Add(k.Valset.Jail(ctx, valAddr, fmt.Sprintf(types.JailReasonNotSupportingTheseExternalChains, strings.Join(notSupported, ", "))))
		}
	}
```

Also, probably also remove this `well well well` comment ;)

deepan95dev: removed the superfluous error check and `well well well`

aleem1314 (approved):

PR #1069, opened 2023-12-28
feat: Upgrade SDK v0.47 to v0.50

deepan95dev:
# Related Github tickets

- Upgrading SDK v0.47.4 to v0.50.2

# Background

_This PR contains all the changes that has to be done for the upgrade, including integration tests/unit tests. This also contains latest master branch commits upto [this](https://github.com/palomachain/paloma/commit/66fb3ff90a08ae1699e068490d4e69f21f622401)_

# Testing completed

- [x] `keeper_integration_tests` are moved to `tests/integration/{module}` with local fixture setup, leaving the dependency on app/app_setup.go (so the file app_setup.go is deleted)
- [x] New test cases added are
* **_util/keeper/address_test.go_** which covers **_util/keeper/address.go_**
* **_/x/gravity/migrations/v1/migrate_test.go_** which covers   **_/x/gravity/migrations/v1/migrate.go_**
- [x] Most of the test coverage exists as it is with upgrade related changes, only few got updated as listed below
* **_tests/integration/evm/keeper/keeper_integration_test.go_** had `putMessageInQueue` in the case `there is another upload valset already in`  only but later we needed to add for two more functions, `TestFirstSnapshot_OnSnapshotBuilt()` function and `TestOldPublishedSnapshot_OnSnapshotBuilt` function
* In **_x/paloma/ante.go_**  after a series of discussion we arrived at the conclusion to use `libmeta.getSigner` but the test cases were failing so I mentioned [here](https://github.com/palomachain/paloma/blob/bcdf4d248c51caa2898f58d4b42e0d91add59395/x/paloma/ante.go#L96) with comment, you may interpret this change to adapt with the test case.
- [x] Upgrade testing is done

# Breaking changes


tests/integration/scheduler/keeper/keeper_test.go
```diff
@@ -0,0 +1,284 @@
...
+	paramskeeper "github.com/cosmos/cosmos-sdk/x/params/keeper"
+	paramstypes "github.com/cosmos/cosmos-sdk/x/params/types"
+	slashingtypes "github.com/cosmos/cosmos-sdk/x/slashing/types"
+	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	"github.com/cosmos/gogoproto/proto"
+	icacontrollertypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/controller/types"
+	icahosttypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/host/types"
+	icatypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/types"
+	ibcfeetypes "github.com/cosmos/ibc-go/v8/modules/apps/29-fee/types"
```
aleem1314, line 41: Use only the required modules, remove unnecessary ones.

tests/integration/valset/keeper/keeper_test.go
```diff
@@ -0,0 +1,284 @@
...
+	paramskeeper "github.com/cosmos/cosmos-sdk/x/params/keeper"
+	paramstypes "github.com/cosmos/cosmos-sdk/x/params/types"
+	paramtypes "github.com/cosmos/cosmos-sdk/x/params/types"
+	slashingkeeper "github.com/cosmos/cosmos-sdk/x/slashing/keeper"
+	slashingtypes "github.com/cosmos/cosmos-sdk/x/slashing/types"
+	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	icacontrollertypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/controller/types"
+	icahosttypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/host/types"
+	icatypes "github.com/cosmos/ibc-go/v8/modules/apps/27-interchain-accounts/types"
```
aleem1314, line 41: ditto here

x/consensus/keeper/concensus_keeper_test.go
```diff
@@ -286,14 +286,14 @@ func TestGettingMessagesThatHaveReachedConsensus(t *testing.T) {
 	}
 	total := math.NewInt(18)
 
-	_ = testValidators
-	_ = total
+	// _ = testValidators
```
aleem1314, line 289: Why this code is commented?

app/app.go
```diff
@@ -441,16 +423,16 @@ func New(
 		app.AccountKeeper,
 		BlockedAddresses(),
 		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
-		logger,
+		log.NewNopLogger(),
```
aleem1314, line 426: Why we need to create new logger?

deepan95dev: reverted

tests/integration/valset/keeper/keeper_test.go
deepan95dev: done

byte-bandit (changes requested):
Blocking this until I find some time to step through all the changes. Other than code review, there's a few things missing here:

- [ ] Rebase on latest version of `master`
- [x] PR description including short description & already performed tests
- [ ] before these changes are added to master, we need full E2E testing including compatibility with Paloma's side loading app Pigeon. We can compile a list of manual testing steps required to pass successfully before we consider going ahead.

byte-bandit:
> How shall we proceed with E2E testing with Paloma's Pigeon App?

@taariq @verabehr @aleem1314 

I suggest Volume compiles a list of manual testing steps and required proof of verification to share with the vitwit team. Happy to drive this on Notion until we settle on a run book.

proto/palomachain/paloma/consensus/tx.proto
```diff
@@ -20,12 +21,14 @@ service Msg {
 }
 
 message MsgAddMessagesSignatures {
+  option (cosmos.msg.v1.signer) = "metadata";
```
byte-bandit, line 24: Is this needed? If so, why - and where can I see the implication of this change in this PR?

Doesn't seem to affect the generated types. According to the docs:
```
option (cosmos.msg.v1.signer) identifies X as the field that will serve as the signer. At compilation time, the SDK will automagically pick up the value of this annotation to have your message implement sdk.Msg.GetSigners.
```

Then again, `metadata` is a composed struct, not an address. Does this change make sense and if so, where can I see it?

byte-bandit (changes requested):
Some questions regarding proto changes.

proto/palomachain/paloma/consensus/tx.proto
deepan95dev: Txns were failing with error related to signer, so we added this to define signer.

proto/palomachain/paloma/gravity/msgs.proto
```diff
@@ -169,3 +174,21 @@ message EventOutgoingTxId {
     string message = 1;
     string tx_id = 2;
 }
+
+message MsgUpdateParams {
+  option (cosmos.msg.v1.signer) = "authority";
```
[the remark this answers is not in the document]
deepan95dev: Yeah, it should be metadata here, will update in the upcoming commit.

util/liblog/liblog.go
```diff
@@ -11,13 +11,18 @@ type Logr interface {
...
 }
 
 type lgwr struct {
 	l log.Logger
 }
 
+// Impl implements Logr.
+func (*lgwr) Impl() any {
+	return log.NewNopLogger()
+}
```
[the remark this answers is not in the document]
deepan95dev: since we use `"cosmossdk.io/log"` we need to implement `Impl()` that returns type `any`. So instead of returning `NopLogger` if we return `nil` also it won't make any difference.

proto/palomachain/paloma/valset/common.proto
```diff
@@ -10,6 +11,7 @@ option go_package = "github.com/palomachain/paloma/x/valset/types";
 // Getters are implemented manually to avoid use of pointer methods, which breaks
 // interface casting.
 message MsgMetadata {
+  option (cosmos.msg.v1.signer) = "signers";
```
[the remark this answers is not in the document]
deepan95dev: Yeah, it should be metadata here, will update in the upcoming commit.

util/liblog/liblog.go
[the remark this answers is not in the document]
byte-bandit: I mean, consumers may still call `Impl()` without checking the underlaying implementation, so `nil` is not a good idea. But since it's just returning any, you'd have to make a type assertion anyway. This is okay I suppose.

proto/palomachain/paloma/consensus/tx.proto
byte-bandit: Can you give the exact error you ran in to? Was it simply enforcing you to specify this information? Because the added options here have not resulting in a change of generated code. So I wonder where this error would have come from and why this seems to fix it.

deepan95dev: [This](https://github.com/cosmos/cosmos-sdk/blob/9a4091478c5e3c21e99670de6eda596ad7d96b29/x/tx/signing/context.go#L160) was the error we got while executing transactions. Since the custom msg type [here](https://github.com/palomachain/paloma/pull/1069/files/bcdf4d248c51caa2898f58d4b42e0d91add59395#diff-e70526f754289da7f9ba89f1c44add9c03d5bb6e3fedcfa7e6b0ec74cfdc7d1bR21) is `metadata`

byte-bandit: When did you encounter the error? During code generation? During compilation? Runtime?

Also, are you saying you encountered this error: `a custom signer function as been defined for message %s which already has a signer field defined with (cosmos.msg.v1.signer)` BEFORE you added the `cosmos.msg.v1.signer` option to our message definitions? That doesn't seem to make sense, as the error will only be thrown if this option is present & a custom signer function exists (which it should).

The reason why I'm pressing so hard on this is because our antehandler which handles our custom solution for using multiple signature keys relies heavily on this `GetSigners()` signature, and we need to ensure that there is no difference in implementation.

deepan95dev: While executing the transaction we got this error.

byte-bandit: Ah, okay. So this is even worse, as it injects the `GetSigner()` function dynamically using reflection...

Which message(s) was/were in that transaction? 

Looking at the implementation [here](https://github.com/cosmos/cosmos-sdk/blob/9a4091478c5e3c21e99670de6eda596ad7d96b29/x/tx/signing/context.go#L159), what happens is:

```
IF cosmos.msg.v1.signer is part of message AND custom getter is defined => Throw error
```

So as long as `cosmos.msg.v1.signer` isn't defined on the messages, this error should not even occur. We don't really want to use dynamic injection of this method, as we need rely on our custom getters.

deepan95dev: Almost all transactions threw the same error, especially  `palomad tx valset add-external-chain-info-for-validator`. Btw the change [here](https://github.com/palomachain/paloma/blob/bcdf4d248c51caa2898f58d4b42e0d91add59395/x/paloma/ante.go#L96) has impact on test cases so I commented it. Does that has any impact on above scenario as well?

byte-bandit: Yes, we need the `libmeta` implementation to work, otherwise we'll run into problems with Pigeons using multiple signing keys.

deepan95dev: Well, we tried out this but ran into test case issues on using `libmeta`. It would help you can look at this `libmeta` usage and the ante_test.go to pass with that usage.

byte-bandit: I can take a look. What does your test setup look like for this? Just `make test`?

atheeshp (commented):
1st round of feedback

proto/palomachain/paloma/consensus/tx.proto
deepan95dev: `make test` would compile all module tests, so specifically to address this `libmeta` issue you might need to run `go test` inside `x/paloma` folder

x/paloma/ante.go
```diff
@@ -90,11 +93,11 @@ func (d VerifyAuthorisedSignatureDecorator) AnteHandle(ctx sdk.Context, tx sdk.T
 		}
 
 		creator := m.GetMetadata().GetCreator()
-		signers := msg.GetSigners()
+		signers := m.GetMetadata().GetSigners() // libmeta.GetSigners(m) //// COMMENTED DUE TO FAILED TEST CASE
```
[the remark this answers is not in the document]
deepan95dev: @byte-bandit  is looking into this issue already.

x/valset/keeper/keeper.go
```diff
@@ -71,32 +76,43 @@ func NewKeeper(
...
+func (k Keeper) Logger(ctx context.Context) log.Logger {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	return sdkCtx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
 
+// TODO: not required now
+func (k Keeper) PunishValidator(ctx context.Context) {}
+
+// TODO: not required now
+func (k Keeper) Heartbeat(ctx context.Context) {}
```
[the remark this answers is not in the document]
deepan95dev: removed these functions

x/scheduler/module.go
```diff
@@ -19,8 +21,20 @@ import (
 )
 
 var (
-	_ module.AppModule      = AppModule{}
-	_ module.AppModuleBasic = AppModuleBasic{}
+	//_ module.AppModule      = AppModule{}
```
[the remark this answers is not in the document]
deepan95dev: removed

x/gravity/types/genesis.go
```diff
@@ -89,172 +85,3 @@ func DefaultGenesisState() *GenesisState {
...
-}
-
-// Equal returns a boolean determining if two Params types are identical.
-func (p Params) Equal(p2 Params) bool {
-	bz1 := ModuleCdc.MustMarshalLengthPrefixed(&p)
-	bz2 := ModuleCdc.MustMarshalLengthPrefixed(&p2)
-	return bytes.Equal(bz1, bz2)
-}
-
-func validateContractHash(i interface{}) error {
```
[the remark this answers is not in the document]
deepan95dev: These were removed after params migration by @vishal-kanna

tests/integration/evm/keeper/keeper_integration_test.go
```diff
@@ -400,41 +420,50 @@ func TestOldPublishedSnapshot_OnSnapshotBuilt(t *testing.T) {
...
 }
 
 func TestInactiveChain_OnSnapshotBuilt(t *testing.T) {
-	a := app.NewTestApp(t, false)
-	ctx := a.NewContext(false, tmproto.Header{
-		Height: 5,
-	})
+	t1 := GinkgoT()
+	f := initFixture(t1)
+	ctx := f.ctx
```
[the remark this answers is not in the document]
deepan95dev: done

tests/integration/evm/keeper/keeper_integration_test.go
```diff
@@ -400,41 +420,50 @@ func TestOldPublishedSnapshot_OnSnapshotBuilt(t *testing.T) {
...
 }
 
 func TestAddingSupportForNewChain(t *testing.T) {
-	a := app.NewTestApp(t, false)
-	ctx := a.NewContext(false, tmproto.Header{
-		Height: 5,
-	})
+	t1 := GinkgoT()
+	f := initFixture(t1)
+	ctx := f.ctx
```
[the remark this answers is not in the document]
deepan95dev: done

tests/integration/evm/keeper/keeper_integration_test.go
```diff
@@ -557,12 +588,12 @@ func TestKeeper_ValidatorSupportsAllChains(t *testing.T) {
...
 					require.NoError(t, err)
 
-					err = a.EvmKeeper.ActivateChainReferenceID(ctx, newChain.ChainReferenceID, &types.SmartContract{Id: 123}, fmt.Sprintf("addr%d", i), []byte("abc"))
+					err = a.evmKeeper.ActivateChainReferenceID(ctx, newChain.ChainReferenceID, &types.SmartContract{Id: 123}, fmt.Sprintf("addr%d", i), []byte("abc"))
 					require.NoError(t, err)
 				}
 
 				validator := genValidators(1, 1000)[0]
-				a.StakingKeeper.SetValidator(ctx, validator)
+				a.stakingKeeper.SetValidator(ctx, validator)
```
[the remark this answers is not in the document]
deepan95dev: done

tests/integration/paloma/keeper/keeper_integration_test.go
```diff
@@ -28,33 +28,29 @@ func TestPalomaGinkgo(t *testing.T) {
...
-
 	BeforeEach(func() {
 		// Generate enough validators to ensure jailing is not prevented due to high
 		// amount of stake in network.
 		vals = testutil.GenValidators(10, 100)
 
 		for _, val := range vals {
-			a.StakingKeeper.SetValidator(ctx, val)
-			a.StakingKeeper.SetValidatorByConsAddr(ctx, val)
+			f.stakingKeeper.SetValidator(ctx, val)
```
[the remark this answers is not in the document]
deepan95dev: done

tests/integration/evm/keeper/keeper_integration_test.go
```diff
@@ -645,14 +678,13 @@ func TestKeeper_ValidatorSupportsAllChains(t *testing.T) {
 	asserter := assert.New(t)
 	for _, tt := range testcases {
 		t.Run(tt.name, func(t *testing.T) {
-			a := app.NewTestApp(t, false)
-			ctx := a.NewContext(false, tmproto.Header{
-				Height: 5,
-			})
+			t1 := GinkgoT()
+			f = initFixture(t1)
+			ctx := f.ctx
```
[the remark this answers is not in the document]
deepan95dev: done

x/gravity/types/genesis.go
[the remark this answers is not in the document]
vishal-kanna: @atheeshp All the params related validations are done in new file `params.go`

x/evm/keeper/grpc_query_params_test.go
```diff
@@ -11,7 +11,7 @@ import (
 
 func TestParamsQuery(t *testing.T) {
 	keeper, ctx := testkeeper.EvmKeeper(t)
-	wctx := sdk.WrapSDKContext(ctx)
+	wctx := sdk.UnwrapSDKContext(ctx)
```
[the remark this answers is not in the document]
vishal-kanna: Done

tests/integration/evm/keeper/keeper_integration_test.go
```diff
@@ -262,33 +275,34 @@ func TestRecentPublishedSnapshot_OnSnapshotBuilt(t *testing.T) {
...
-	err = a.ValsetKeeper.SaveModifiedSnapshot(ctx, latestSnapshot)
+	err = f.valsetKeeper.SaveModifiedSnapshot(ctx, latestSnapshot)
 	require.NoError(t, err)
 
 	// Add two validators to make this new snapshot worthy
 	validators = genValidators(2, 25000)
 	for _, val := range validators {
-		a.StakingKeeper.SetValidator(ctx, val)
-		err = a.ValsetKeeper.AddExternalChainInfo(ctx, val.GetOperator(), []*valsettypes.ExternalChainInfo{
+		f.stakingKeeper.SetValidator(ctx, val)
```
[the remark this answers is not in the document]
vishal-kanna: Handled the error

tests/integration/evm/keeper/keeper_integration_test.go
```diff
@@ -199,34 +205,40 @@ func TestFirstSnapshot_OnSnapshotBuilt(t *testing.T) {
...
 
 	queue := fmt.Sprintf("evm/%s/%s", newChain.GetChainReferenceID(), keeper.ConsensusTurnstoneMessage)
-
-	msgs, err := a.ConsensusKeeper.GetMessagesFromQueue(ctx, queue, 100)
+	msgs, err := f.consensusKeeper.GetMessagesFromQueue(ctx, queue, 100)
 	require.NoError(t, err)
 	require.Empty(t, msgs)
-
-	_, err = a.ValsetKeeper.TriggerSnapshotBuild(ctx)
+	f.consensusKeeper.PutMessageInQueue(ctx, queue, &types.Message{
```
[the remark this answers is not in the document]
vishal-kanna: Handled the error

tests/integration/evm/keeper/keeper_integration_test.go
```diff
@@ -186,8 +191,9 @@ func TestFirstSnapshot_OnSnapshotBuilt(t *testing.T) {
 
 	validators := genValidators(25, 25000)
 	for _, val := range validators {
-		a.StakingKeeper.SetValidator(ctx, val)
-		err = a.ValsetKeeper.AddExternalChainInfo(ctx, val.GetOperator(), []*valsettypes.ExternalChainInfo{
+		f.stakingKeeper.SetValidator(ctx, val)
```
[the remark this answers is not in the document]
vishal-kanna: Handled the error

tests/integration/evm/keeper/keeper_integration_test.go
```diff
@@ -63,35 +62,37 @@ func TestEndToEndForEvmArbitraryCall(t *testing.T) {
...
 		require.NoError(t, err)
-		err = a.ValsetKeeper.AddExternalChainInfo(ctx, validator.GetOperator(), []*valsettypes.ExternalChainInfo{
+		operator, err := utilkeeper.ValAddressFromBech32(f.evmKeeper.AddressCodec, validator.GetOperator())
+		require.NoError(t, err)
+		err = f.valsetKeeper.AddExternalChainInfo(ctx, operator, []*valsettypes.ExternalChainInfo{
 			{
 				ChainType:        "evm",
 				ChainReferenceID: newChain.GetChainReferenceID(),
-				Address:          valAddr.String(),
+				Address:          string(valAddr),
```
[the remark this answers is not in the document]
vishal-kanna: Used the address codec to convert into the string

x/valset/keeper/keeper.go
```diff
@@ -525,13 +555,18 @@ func (k Keeper) GetSigningKey(ctx sdk.Context, valAddr sdk.ValAddress, chainType
...
 
 // IsJailed returns if the current validator is jailed or not.
-func (k Keeper) IsJailed(ctx sdk.Context, val sdk.ValAddress) bool {
-	return k.staking.Validator(ctx, val).IsJailed()
+func (k Keeper) IsJailed(ctx context.Context, val sdk.ValAddress) (bool, error) {
+	a, err := k.staking.Validator(ctx, val)
+	if err != nil {
+		return a.IsJailed(), err
+	}
+	return a.IsJailed(), nil
```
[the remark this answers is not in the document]
deepan95dev: This has been used in multiple files and test cases also required by `mocks.StakingKeeper`

x/gravity/module.go
```diff
@@ -22,8 +24,20 @@ import (
 // type check to ensure the interface is properly implemented
 // nolint: exhaustruct
 var (
-	_ module.AppModule      = AppModule{}
-	_ module.AppModuleBasic = AppModuleBasic{}
+	// _ module.AppModule           = AppModule{}
```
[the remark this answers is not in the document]
deepan95dev: done

x/gravity/migrations/v1/migrate.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"cosmossdk.io/core/store"
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/palomachain/paloma/x/gravity/exported"
+	"github.com/palomachain/paloma/x/gravity/types"
+)
+
+func MigrateParams(
```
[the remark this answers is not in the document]
deepan95dev: done

x/scheduler/keeper/params.go
```diff
@@ -1,16 +1,15 @@
...
 import (
-	sdk "github.com/cosmos/cosmos-sdk/types"
+	"context"
+
 	"github.com/palomachain/paloma/x/scheduler/types"
 )
 
-// GetParams get all parameters as types.Params
-func (k Keeper) GetParams(ctx sdk.Context) types.Params {
+func (k Keeper) GetParams(ctx context.Context) (params types.Params) {
```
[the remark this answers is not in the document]
deepan95dev: done

x/valset/keeper/keeper.go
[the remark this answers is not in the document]
deepan95dev: This is being used by other functions and test cases, also required by interface `types.valsetKeeper`

proto/palomachain/paloma/consensus/tx.proto
byte-bandit: @deepan95dev I made the necessary change and all tests are green. I'm not sure where this error would come in that you spoke of? I pushed the changes to the branch, please can you try and reproduce this error?

deepan95dev: sure, I will try to attach the error screenshot in txn, to address the issue better

deepan95dev: Here is the error message if we remove the "metadata" signer   @byte-bandit  @atheeshp 
TXN_ERROR

byte-bandit: Okay, `no cosmos.msg.v1.siger option found`. That's a different error than you linked me to here: https://github.com/palomachain/paloma/pull/1069#discussion_r1462733799

It makes sense. The code is trying to figure out which property represents the signing authority for the message, now that the official `GetSigners()` method was removed.

What happens when you use `DefineCustomGetSigners()` as hinted at by the error log?

---

Just FYI, I also spent a few hours today running the latest version of this branch and the changed Pigeon on a private test net, and there seem to be some changes with regards to setup and configuration as well. 

The last error I ran into was `invalid coin denomination: got ugrain, expected stake`, at which I stopped for now. Just an FYI that there's more issues ahead after resolving this.

deepan95dev: Yeah, I should have pointed [this error message](https://github.com/cosmos/cosmos-sdk/blob/d8fb76d4c2671d1bd8ab656a66752091589bb4ad/x/tx/signing/context.go#L119) in above [#1069 (comment)](https://github.com/palomachain/paloma/pull/1069#discussion_r1462733799).

So this error message was fixed with `customGetSigners` [here](https://github.com/palomachain/paloma/blob/481fc23e60ec72416b23381556849894ff7975b2/app/app.go#L308)  which requires `cosmos.msg.v1.siger option` in proto files, so we need to revert [this commit](https://github.com/palomachain/paloma/pull/1069/commits/481fc23e60ec72416b23381556849894ff7975b2). 


Also can you list down the steps to test along with Pigeon App?

deepan95dev: > The last error I ran into was `invalid coin denomination: got ugrain, expected stake`, at which I stopped for now. Just an FYI that there's more issues ahead after resolving this.

For this issue we used `--default-denom ugrain` flag while initiating genesis.

byte-bandit: Please read up on these changes, this is an important change for us and it needs to be well understood. The `cosmos.msg.v1.signer` option is the default CosmosSDK implementation on how to extract signing information from a message. It is needed because `GetSigners()` is no longer part of the `Msg` interface. 

The background of this is easy. With the new option, retrieving the signers is automated. You no longer have to define it yourself for every message, which in theory is great.

However, Paloma uses multiple fields for signature information and aggregates those into one flattened list of signers. The arithmetic logic for this is in `libmeta`. That's why it is **not** enough to simply read one property of the message.

The default `cosmos.msg.v1.signer` simply extracts tells the framework which field to extract when looking at signing infos. This will **not** work with our implementation, because - again - Paloma needs to inspect both the `Creator` **and** the `Signers` field on the metadata.

Luckily, we're not the only ones with this problem, i.e. see [here](https://github.com/cosmos/cosmos-sdk/issues/16112). So Cosmos has introduced this `customGetSigners` concept which allows you to specify your own functions for signer identification, and override the `cosmos.msg.v1.signer` implementation.

Specifying `cosmos.msg.v1.signer` with `customGetSigners` as you have done is taking the whole idea of it Ad Absurdum. Instead, you want to feed in your own retriever function, which again makes use of `libmeta`.

Please also do your own research into this.

---

As for the testing alongside Pigeon, we have a full run book for you, but I think it makes sense to fix all of these issues before we kick off E2E testing.

deepan95dev:
@byte-bandit we have implemented `CustomGetSigners` and moved the `InterfaceRegistry` to app/interface_registry.go which comprises all MetaData signers info.

byte-bandit:
> @byte-bandit we have implemented `CustomGetSigners` and moved the `InterfaceRegistry` to app/interface_registry.go which comprises all MetaData signers info.

Great. Let's keep this in the history. 

I think it's actually worth trying to utilise the `cosmos.msg.v1.signer` option instead of the custom getter. I thought that the `libmeta` implementation did more than just return the `[]signers` field, but it doesn't. The combinatory logic only happens in the ante handler.

Apologies for the confusion. Remove the custom getter (though keep it in git history please) and re-introduce the `cosmos.msg.v1.signer` options pointing to the metadata field on all messages please.

vishal-kanna:
Okay

x/gravity/keeper/evidence.go
```diff
@@ -82,25 +83,31 @@ func (k Keeper) checkBadSignatureEvidenceInternal(ctx sdk.Context, subject types
...
 	}
-
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	params := k.GetParams(ctx)
 	if !val.IsJailed() {
-		k.StakingKeeper.Jail(ctx, cons)
-		k.StakingKeeper.Slash(ctx, cons, ctx.BlockHeight(), val.ConsensusPower(sdk.DefaultPowerReduction), params.SlashFractionBadEthSignature)
+		err := k.StakingKeeper.Jail(ctx, cons)
+		if err != nil {
+			return err
```
byte-bandit, line 91:
```suggestion
			return fmt.Errorf("checkBadSignatureEvidenceInternal jail: %w", err)
```

x/gravity/keeper/keeper.go
```diff
@@ -29,45 +30,38 @@ var (
...
-
-	storeGetter keeperutil.StoreGetter
+	AddressCodec      address.Codec
+	storeGetter       keeperutil.StoreGetter
 
 	AttestationHandler interface {
-		Handle(sdk.Context, types.Attestation, types.EthereumClaim) error
+		Handle(context.Context, types.Attestation, types.EthereumClaim) error
 	}
+	authority string
```
byte-bandit, line 47: I don't think Paloma uses any authority gated features. Is this implementation enforced by Cosmos?

byte-bandit: So, params updates on Paloma are currently all done by governance proposals, and our plans are to keep it this way, allowing anyone in the community to propose params udpates.

This `authority` concept is new to Cosmos I believe, and from what I can tell, an update to the params is now a simple message, but must originate from the authority address?

Can we ignore this approach and stick with proposals only?

x/gravity/keeper/test_common.go
```diff
@@ -54,22 +59,26 @@ import (
...
+	capabilitykeeper "github.com/cosmos/ibc-go/modules/capability/keeper"
+	capabilitytypes "github.com/cosmos/ibc-go/modules/capability/types"
+	ibctransferkeeper "github.com/cosmos/ibc-go/v8/modules/apps/transfer/keeper"
+	ibctransfertypes "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	ibcexported "github.com/cosmos/ibc-go/v8/modules/core/exported"
+	ibckeeper "github.com/cosmos/ibc-go/v8/modules/core/keeper"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
-	gravityparams "github.com/palomachain/paloma/app/params"
+	params2 "github.com/palomachain/paloma/app/params"
```
byte-bandit, line 71:
```suggestion
	"github.com/palomachain/paloma/app/params"
```

x/gravity/keeper/test_common.go
```diff
@@ -291,27 +298,32 @@ func addValidators(t *testing.T, input *TestInput, count int) {
...
-		require.True(t, found)
+		require.NoError(t, found)
 
 		pubKey, err := validator.ConsPubKey()
 		require.NoError(t, err)
-
-		err = input.ValsetKeeper.AddExternalChainInfo(input.Context, validator.GetOperator(), []*valsettypes.ExternalChainInfo{
+		valAddress, err := utilkeeper.ValAddressFromBech32(input.GravityKeeper.AddressCodec, validator.GetOperator())
+		if err != nil {
+			input.GravityKeeper.Logger(input.Context).Error("error while converting into bytes", err)
```
byte-bandit, line 323: Why are we logging in tests?

x/gravity/keeper/test_common.go
```diff
@@ -635,43 +671,53 @@ func CreateTestEnv(t *testing.T) TestInput {
...
+		scopedTransferKeeper,
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
 	)
 
 	valsetKeeper := valsetkeeper.NewKeeper(
 		marshaler,
-		keyValset,
-		memKeyValset,
+		runtime.NewKVStoreService(keyValset),
+		// memKeyValset,
```
byte-bandit, line 694: Remove commented code.

x/gravity/keeper/test_common.go
```diff
@@ -635,43 +671,53 @@ func CreateTestEnv(t *testing.T) TestInput {
...
+		authcodec.NewBech32Codec(params2.ValidatorAddressPrefix),
 	)
 
 	consensusRegistry := consensuskeeper.NewRegistry()
 	consensusKeeper := consensuskeeper.NewKeeper(
 		marshaler,
-		keyConsensus,
-		memKeyConsensus,
+		runtime.NewKVStoreService(keyConsensus),
+		// memKeyConsensus,
```
byte-bandit, line 707: dito

x/gravity/keeper/test_common.go
```diff
@@ -635,43 +671,53 @@ func CreateTestEnv(t *testing.T) TestInput {
...
 		consensusRegistry,
 	)
 
 	evmKeeper := evmkeeper.NewKeeper(
 		marshaler,
-		keyEvm,
-		memKeyEvm,
-		getSubspace(paramsKeeper, evmtypes.ModuleName),
+		runtime.NewKVStoreService(keyEvm),
+		// memKeyEvm,
```
byte-bandit, line 716: here and every other occurrence.

x/gravity/types/msgs.go
```diff
@@ -329,3 +331,26 @@ func (msg MsgSubmitBadSignatureEvidence) Type() string { return "Submit_Bad_Sign
...
+
+// ValidateBasic performs stateless checks
+func (msg *MsgUpdateParams) ValidateBasic() (err error) {
+	return libmeta.ValidateBasic(msg)
+}
+
+// GetSignBytes encodes the message for signing
+func (msg MsgUpdateParams) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(msg))
+}
```
byte-bandit, line 345: I think GetSignBytes() is no longer needed?

x/metrix/keeper/keeper.go
```diff
@@ -186,6 +194,11 @@ func (k Keeper) OnSnapshotBuilt(ctx sdk.Context, snapshot *valsettypes.Snapshot)
 			logger.Info("Skip updating metrics, no chains found.")
 			return
 		}
+		scoreMax = int64(len(v.ExternalChainInfos))
+		if scoreMax < 1 {
+			logger.Info("Skip updating metrics, no chains found.")
+			return
+		}
```
byte-bandit, line 201: This should already be on the current master, looks like you're not fully rebased?

x/valset/keeper/grpc_query_get_alive_pigeons.go
```diff
@@ -21,10 +22,15 @@ func (k Keeper) GetAlivePigeons(goCtx context.Context, req *types.QueryGetAliveP
...
 	res := slice.Map(vals, func(val stakingtypes.ValidatorI) *types.QueryGetAlivePigeonsResponse_ValidatorAlive {
-		until, err := k.ValidatorAliveUntil(ctx, val.GetOperator())
+		bz, err := keeperutil.ValAddressFromBech32(k.AddressCodec, val.GetOperator())
+
 		s := &types.QueryGetAlivePigeonsResponse_ValidatorAlive{
-			ValAddress: val.GetOperator(),
+			ValAddress: bz,
+		}
+		if err != nil {
+			s.Error = err.Error()
```
byte-bandit, line 31: Move error handling to after assignment, and what should happen in case we actually encounter an error here? probably we do not just want to continue with regular logic flow?

x/valset/keeper/keep_alive.go
```diff
@@ -128,28 +142,34 @@ func (k Keeper) UpdateGracePeriod(ctx sdk.Context) {
...
 
-func (k Keeper) JailInactiveValidators(ctx sdk.Context) error {
+func (k Keeper) JailInactiveValidators(ctx context.Context) error {
 	var g whoops.Group
 	for _, val := range k.GetUnjailedValidators(ctx) {
 		if !(val.GetStatus() == stakingtypes.Bonded || val.GetStatus() == stakingtypes.Unbonding) {
 			continue
 		}
-		valAddr := val.GetOperator()
+		valAddr, err := keeperutil.ValAddressFromBech32(k.AddressCodec, val.GetOperator())
```
byte-bandit, line 171: error handling

x/valset/keeper/keeper.go
```diff
@@ -374,37 +384,48 @@ func (k Keeper) ValidatorSupportsAllChains(ctx sdk.Context, validatorAddress sdk
...
-func (k Keeper) createNewSnapshot(ctx sdk.Context) (*types.Snapshot, error) {
+func (k Keeper) createNewSnapshot(ctx context.Context) (*types.Snapshot, error) {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	validators := []stakingtypes.ValidatorI{}
-	k.staking.IterateValidators(ctx, func(_ int64, val stakingtypes.ValidatorI) bool {
-		if val.IsBonded() && !val.IsJailed() && k.ValidatorSupportsAllChains(ctx, val.GetOperator()) {
+	err := k.staking.IterateValidators(ctx, func(_ int64, val stakingtypes.ValidatorI) bool {
+		bz, err := keeperutil.ValAddressFromBech32(k.AddressCodec, val.GetOperator())
+		if err != nil {
+			return false
```
byte-bandit, line 399: Log error

x/valset/keeper/keeper.go
```diff
@@ -525,13 +549,18 @@ func (k Keeper) GetSigningKey(ctx sdk.Context, valAddr sdk.ValAddress, chainType
...
+	}
+	return a.IsJailed(), nil
 }
 
-func (k Keeper) Jail(ctx sdk.Context, valAddr sdk.ValAddress, reason string) error {
+func (k Keeper) Jail(_ctx context.Context, valAddr sdk.ValAddress, reason string) error {
+	ctx := sdk.UnwrapSDKContext(_ctx)
 	jailTime := ctx.BlockTime()
-	val := k.staking.Validator(ctx, valAddr)
+	val, err := k.staking.Validator(ctx, valAddr)
```
byte-bandit, line 563: error handling

x/valset/keeper/keeper.go
```diff
@@ -673,26 +712,19 @@ func (k Keeper) externalChainInfoStore(ctx sdk.Context, val sdk.ValAddress) sdk.
...
+	store := runtime.KVStoreAdapter(k.storeKey.OpenKVStore(ctx))
+	return prefix.NewStore(store, []byte("external-chain-info"))
 }
 
-func (k Keeper) snapshotStore(ctx sdk.Context) sdk.KVStore {
-	k.Logger(ctx).Debug("snapshot store", "store-key-name", k.storeKey.Name(), "store-key-string", k.storeKey.String())
-	return prefix.NewStore(ctx.KVStore(k.storeKey), []byte("snapshot"))
+func (k Keeper) snapshotStore(ctx context.Context) storetypes.KVStore {
+	store := runtime.KVStoreAdapter(k.storeKey.OpenKVStore(ctx))
+	// k.Logger(ctx).Debug("snapshot store", "store-key-name", k.memKey.Name(), "store-key-string", k.storeKey) //TODO
```
byte-bandit, line 722: TODO?

x/valset/keeper/keeper_test.go
```diff
@@ -3,7 +3,8 @@ package keeper
 import (
```
byte-bandit, line 3: Add test cases for the new error logic flows.

x/valset/keeper/setup_test.go
```diff
@@ -24,12 +28,18 @@ type mockedServices struct {
 	SlashingKeeper *mocks.SlashingKeeper
 }
 
-func newValsetKeeper(t testing.TB) (*Keeper, mockedServices, sdk.Context) {
-	storeKey := sdk.NewKVStoreKey(types.StoreKey)
+func newValsetKeeper(t testing.TB) (*Keeper, mockedServices, context.Context) {
+	config := sdk.GetConfig()
+	config.SetBech32PrefixForAccount("paloma", "pub")
+	config.SetBech32PrefixForValidator("palomavaloper", "valoperpub")
```
byte-bandit, line 34: use `CreateTestEnv`

byte-bandit (changes requested):
There's a lot of noise coming from the change in context. Since almost the entire code base is reliant on the `sdk.Context` implementation anyway, we'd be able to save a lot of touches by using `func(goCtx context.Context) {ctx := sdk.UnwrapContext(goCtx)}` instead of going the other way.  

That would also render the remaining clode base more legible and idiomatic, and prevent bugs introduced by mass "search/replace" tools where `ctx` is overriden with a scoped version.

Though going forward, it's probably better to stay in line with the Cosmos style of `sdkCtx`.

x/valset/keeper/setup_test.go
vishal-kanna: `CreateTestEnv` creates Testing environment but here we need only above config lines.So that is reason we have added here.

x/valset/keeper/keeper.go
vishal-kanna: We have removed the `memkey` storage because we are not using it anywhere ,so here in logger we have `memkey.name` and we need to remove it .That's why we have added TODO here.

x/valset/keeper/keeper_test.go
vishal-kanna: Added a error logic .Can you look into it

vishal-kanna:
> There's a lot of noise coming from the change in context. Since almost the entire code base is reliant on the `sdk.Context` implementation anyway, we'd be able to save a lot of touches by using `func(goCtx context.Context) {ctx := sdk.UnwrapContext(goCtx)}` instead of going the other way.
> 
> That would also render the remaining clode base more legible and idiomatic, and prevent bugs introduced by mass "search/replace" tools where `ctx` is overriden with a scoped version.
> 
> Though going forward, it's probably better to stay in line with the Cosmos style of `sdkCtx`.
 @byte-bandit  Is it adding a function and using it inplace of the other way

byte-bandit:
> > There's a lot of noise coming from the change in context. Since almost the entire code base is reliant on the `sdk.Context` implementation anyway, we'd be able to save a lot of touches by using `func(goCtx context.Context) {ctx := sdk.UnwrapContext(goCtx)}` instead of going the other way.
> > That would also render the remaining clode base more legible and idiomatic, and prevent bugs introduced by mass "search/replace" tools where `ctx` is overriden with a scoped version.
> > Though going forward, it's probably better to stay in line with the Cosmos style of `sdkCtx`.
> > @byte-bandit  Is it adding a function and using it inplace of the other way

That's not what I meant. There's absolutely nothing we gain from simply replacing`sdkCtx := sdk.UnwrapSDKContext` with with `sdkCtx := common.SdkContext`.

I was talking about function signatures and how we are naming the parameters. But I'm not going to go into detail here now, as the SDK itself uses this naming paradigm, and I believe we should stick with it - even though I personally think it's a subpar solution. We just need to be very careful now, that we actually use `sdkCtx` everywhere after it was cast, and respect branching contexts as well. 

So, please revert back to `sdk.UnwrapSDKContext`.

app/interface_registry.go
```diff
@@ -0,0 +1,100 @@
...
+	if InterfaceRegistry, err = NewInterfaceRegistry(
+		params2.AccountAddressPrefix,
+		params2.ValidatorAddressPrefix,
+	); err != nil {
+		panic(err)
+	}
+}
+
+// getCustomMsgSignerFn returns CustomSigners
+func getCustomMsgSignerFn(a address.Codec, path []string) func(msg golangproto.Message) ([][]byte, error) {
```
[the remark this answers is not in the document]
byte-bandit: I meant keep the change in git history, don't rebase it out of there. It doesn't need to stay in the active code base. Please remove.

proto/palomachain/paloma/consensus/tx.proto
```diff
@@ -35,6 +37,7 @@ message ConsensusMessageSignature {
 message MsgAddMessagesSignaturesResponse {}
 
 message MsgDeleteJob {
+  option (cosmos.msg.v1.signer) = "metadata";
   string                                creator        = 1 [deprecated = true];
```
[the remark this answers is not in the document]
byte-bandit: Yeah, that's expected, there's still some usages in code, but this is the best place to remove them we're going to get. So, remove `creator|signer|authority` from this message (and any other message definitions). This information should live in `Metadata.Creator`, which is a shared field across all messages. So lets remove the specific implementations from the messages, and where there are usages in code that access them, just relink them to `Metadata.Creator`.

x/valset/keeper/setup_test.go
byte-bandit: Suggestion: let's move this definition to a custom function then. `SetupPalomaPrefixes` for example. It can then be used from within `CreateTestEnv`, but also called directly from here.

x/evm/keeper/attest_validator_balances.go
```diff
@@ -96,7 +98,17 @@ func (k Keeper) processValidatorBalanceProof(
 				)
 			}
 			if balance.Cmp(minBalance) == -1 || balance.Cmp(big.NewInt(0)) == 0 {
-				if !k.Valset.IsJailed(ctx, valAddr) {
+				isJailed, err := k.Valset.IsJailed(ctx, valAddr)
+				if err != nil {
+					k.Logger(ctx).Error(
```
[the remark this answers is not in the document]
byte-bandit: Take a look at the actual implementation and compare what happens. The output of ```liblog.FromSDKLogger().
WithError(err).
WithValidator(valAddr).
WithFields(..).
Error("attestValidatorBalances: error in checking jailed validator")` ``

is quite different to:

```
k.Logger(ctx).Error(
					"error setting validator balance",
					"err", err,
					"val-addr", valAddr,
					"eth-addr", hexAddr,
				)
```

Most importantly around the log fields. We don't want specific implementations for these by individual engineers which result in ambiguous field names (`err`, `error`, `E` and so ...). We want one holistic field that we can create a log span for and index appropriately.

x/gravity/keeper/keeper.go
byte-bandit: Can you go a bit into detail on this? I'm not aware of this concept on Paloma as of right now. If it was introduced with Cosmos SDK `0.50`, I'm not sure we want to adopt it.

vishal-kanna: https://docs.cosmos.network/v0.50/build/architecture/adr-046-module-params this ADR might provide some info about it

proto/palomachain/paloma/consensus/tx.proto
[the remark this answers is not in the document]
vishal-kanna: removed the depreciated creator|signer|sender and added `Metadata.Creator`

x/paloma/ante.go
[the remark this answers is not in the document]
vishal-kanna: > @byte-bandit is looking into this issue already.
@byte-bandit  Do you have suggestion on this?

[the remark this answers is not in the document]
byte-bandit: I didn't run into any failing tests when I used `libmeta.GetSigners()`, it's possible this is no longer an issue with the added `signer` options defined on the proto files?

x/valset/keeper/keeper.go
byte-bandit: Resolve the TODO.

app/interface_registry.go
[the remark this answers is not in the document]
byte-bandit: Not sure why you resolved this? The function is still here?

proto/palomachain/paloma/evm/tx.proto
```diff
@@ -12,6 +12,7 @@ service Msg {
 }
 
 message MsgDeployNewSmartContractRequest {
+  option (cosmos.msg.v1.signer) = "metadata";
```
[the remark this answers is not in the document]
byte-bandit: It's still here?

proto/palomachain/paloma/evm/tx.proto
```diff
@@ -24,10 +25,12 @@ message MsgDeployNewSmartContractRequest {
 message DeployNewSmartContractResponse {}
 
 message MsgRemoveSmartContractDeploymentRequest {
+  option (cosmos.msg.v1.signer) = "metadata";
```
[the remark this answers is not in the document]
byte-bandit: Still here.

x/valset/keeper/keeper.go
```diff
@@ -673,26 +714,19 @@ func (k Keeper) externalChainInfoStore(ctx sdk.Context, val sdk.ValAddress) sdk.
...
+	store := runtime.KVStoreAdapter(k.storeKey.OpenKVStore(ctx))
+	return prefix.NewStore(store, []byte("external-chain-info"))
 }
 
-func (k Keeper) snapshotStore(ctx sdk.Context) sdk.KVStore {
-	k.Logger(ctx).Debug("snapshot store", "store-key-name", k.storeKey.Name(), "store-key-string", k.storeKey.String())
-	return prefix.NewStore(ctx.KVStore(k.storeKey), []byte("snapshot"))
+func (k Keeper) snapshotStore(ctx context.Context) storetypes.KVStore {
+	store := runtime.KVStoreAdapter(k.storeKey.OpenKVStore(ctx))
+	k.Logger(ctx).Debug("snapshot store", "store-key-name", "store-key-string", k.storeKey)
```
byte-bandit, line 724: Why would you leave the field key and only remove value? Now the whole ordering doesn't work anymore.

Just remove the entire statement, I don't think it's needed.

x/valset/keeper/keeper.go
```diff
@@ -525,14 +549,21 @@ func (k Keeper) GetSigningKey(ctx sdk.Context, valAddr sdk.ValAddress, chainType
...
 
-func (k Keeper) Jail(ctx sdk.Context, valAddr sdk.ValAddress, reason string) error {
+func (k Keeper) Jail(_ctx context.Context, valAddr sdk.ValAddress, reason string) error {
+	ctx := sdk.UnwrapSDKContext(_ctx)
 	jailTime := ctx.BlockTime()
-	val := k.staking.Validator(ctx, valAddr)
-
+	val, err := k.staking.Validator(ctx, valAddr)
+	if err != nil {
+		return ErrValidatorWithAddrNotFound.Format(valAddr)
```
byte-bandit, line 565: Why would you not return the actual error?

app/interface_registry.go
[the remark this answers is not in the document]
vishal-kanna: Will be removed

x/valset/keeper/keeper.go
vishal-kanna: ok

x/valset/keeper/keeper.go
vishal-kanna: Actual `err` and `ErrValidatorWithAddrNotFound` specify the same thing ( i.e validator does not exist).So i used the ErrValidatorWithAddrNotFound error instead of actual error

x/paloma/ante.go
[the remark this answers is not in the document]
vishal-kanna: I can't see any issue with proto fields.

x/valset/keeper/keeper.go
vishal-kanna: Resolved

x/gravity/keeper/keeper.go
byte-bandit: It doesn't mention this authority concept. Also, it doesn't appear to have gone past the proposal stage.

x/valset/keeper/keeper.go
vishal-kanna: Okay
I've returned the actual error.

vishal-kanna: @byte-bandit Can you resolve the completed conversation

x/valset/keeper/keeper.go
```diff
@@ -525,14 +549,21 @@ func (k Keeper) GetSigningKey(ctx sdk.Context, valAddr sdk.ValAddress, chainType
...
 
-func (k Keeper) Jail(ctx sdk.Context, valAddr sdk.ValAddress, reason string) error {
+func (k Keeper) Jail(_ctx context.Context, valAddr sdk.ValAddress, reason string) error {
+	ctx := sdk.UnwrapSDKContext(_ctx)
 	jailTime := ctx.BlockTime()
-	val := k.staking.Validator(ctx, valAddr)
-
+	val, err := k.staking.Validator(ctx, valAddr)
+	if err != nil {
+		return whoops.Errorf(err.Error()).Format()
```
byte-bandit, line 565: Can you explain what exactly you're doing here - and why?

vishal-kanna: If i return the `error` ,i'm getting an error from the integration test. So i have used this error
And all the below return use the `whoops.Errof`

byte-bandit: We're trying to remove this `whoops` library. But it's out of scope for this. Thanks for clearing it up.

vishal-kanna: I have returned the actual error now

x/evm/keeper/attest.go
```diff
@@ -109,15 +112,15 @@ func (k Keeper) attestRouter(ctx sdk.Context, q consensus.Queuer, msg consensust
...
 			}
 		}
 
 		handledAt := msg.GetHandledAtBlockHeight()
 		if handledAt == nil {
-			handledAt = func(i math.Int) *math.Int { return &i }(sdk.NewInt(ctx.BlockHeight()))
+			handledAt = func(i math.Int) *math.Int { return &i }(sdkmath.NewInt(ctx1.BlockHeight()))
 		}
-		publishMessageAttestedEvent(ctx, &k, msg.GetId(), message.Assignee, message.AssignedAtBlockHeight, *handledAt, success)
+		publishMessageAttestedEvent(sdkCtx, &k, msg.GetId(), message.Assignee, message.AssignedAtBlockHeight, *handledAt, success)
```
byte-bandit, line 123: This needs to be `ctx1` as well I think.

x/evm/keeper/attest.go
```diff
@@ -71,8 +75,7 @@ func (k Keeper) attestRouter(ctx sdk.Context, q consensus.Queuer, msg consensust
 		return nil
 	}
 
-	// Use transactionalsed context only!
-	ctx, writeCache := ctx.CacheContext()
+	ctx1, writeCache := sdkCtx.CacheContext()
```
byte-bandit, line 78:
```suggestion
	cacheCtx, writeCache := sdkCtx.CacheContext()
```

x/gravity/keeper/keeper.go
byte-bandit: Hmm, maybe the ADR is outdated. The official docs mention it. I suppose let's take the benefit.

x/gravity/types/msgs.go
```diff
@@ -329,3 +328,21 @@ func (msg MsgSubmitBadSignatureEvidence) Type() string { return "Submit_Bad_Sign
...
+
+// ValidateBasic performs stateless checks
+func (msg *MsgUpdateParams) ValidateBasic() (err error) {
+	return libmeta.ValidateBasic(msg)
+}
+
+// GetSigners defines whose signature is required
+func (msg *MsgUpdateParams) GetSigners() []sdk.AccAddress {
+	return libmeta.GetSigners(msg)
+}
```
byte-bandit, line 342: This can be removed.

x/metrix/keeper/keeper.go
```diff
@@ -361,12 +384,18 @@ func (k *Keeper) UpdateUptime(ctx sdk.Context) {
 			WithFields(
 				"missed-blocks-counter", info.MissedBlocksCounter,
 				"uptime", uptime,
+				"is-jailed", isJailed,
+				"uptime", uptime,
```
byte-bandit, line 388: Copy paste mistake?

x/metrix/module.go
```diff
@@ -153,16 +166,20 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
...
 // returns no validator updates.
-func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
-	if ctx.BlockHeight()%cUpdateUptimeBlockInterval == 0 {
+func (am AppModule) EndBlock(ctx context.Context) error {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	if sdkCtx.BlockHeight()%cUpdateUptimeBlockInterval == 0 {
 		am.keeper.PurgeRelayMetrics(ctx)
 		am.keeper.UpdateRelayMetrics(ctx)
-		am.keeper.UpdateUptime(ctx)
+		err := am.keeper.UpdateUptime(ctx)
```
byte-bandit, line 178: I don't think this needs to return an error. The current definition doesn't, and this is really just a fire & forget mechanism to track what's been going on. Logging the errors in `UpdateUpdate` is enough, we dont' need to bubble them up here.

byte-bandit (changes requested):
Only small stuff remaining, but most of it we'll just accept for now. 

Can you please also run a full scan over the code base and make sure there's no other occurrances of `CacheContext()` that slipped through the renaming?

x/evm/keeper/attest.go
vishal-kanna: okay

x/metrix/module.go
vishal-kanna: Okay

x/metrix/keeper/keeper.go
```diff
@@ -344,11 +344,11 @@ func (k *Keeper) UpdateRelayMetrics(ctx context.Context) {
 	logger.Debug("Updating relay metrics finished!")
 }
 
-func (k *Keeper) UpdateUptime(ctx context.Context) error {
+func (k *Keeper) UpdateUptime(ctx context.Context) {
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 	window, err := k.slashing.SignedBlocksWindow(ctx)
 	if err != nil {
-		return err
+		k.Logger(sdkCtx).With("Err", err).Error("Error while getting the SignedBlocksWindow")
```
byte-bandit, line 351: What did we learn about the difference of doing this vs. using liblog?

vishal-kanna: `K.Logger` and  `liblog` are different .They have different usage.

byte-bandit (approved):
Final review to @aleem1314

app/app.go
```diff
@@ -476,13 +469,14 @@ func New(
 		app.StakingKeeper,
 		app.UpgradeKeeper,
 		scopedIBCKeeper,
+		authtypes.NewModuleAddress(govtypes.ModuleName).String(),
```
aleem1314, line 472: this is repeated multiple times. Create a const for this.

x/gravity/migrations/v1/migrate.go
```diff
@@ -0,0 +1,30 @@
...
+	"cosmossdk.io/core/store"
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/palomachain/paloma/x/gravity/exported"
+	"github.com/palomachain/paloma/x/gravity/types"
+)
+
+// migrateParams will set the params to store from subspace
+
+func MigrateParams(
```
aleem1314, line 13: Have you manually tested if the migration function is being executed? I don't see a consensus version bump for this migration logic in the `module.go` file.

vishal-kanna: No,i didn't test manually.But i now have added the consensus version bump and checked it

go.mod
```diff
@@ -4,225 +4,244 @@ go 1.21
...
-	github.com/cometbft/cometbft v0.37.2
-	github.com/cometbft/cometbft-db v0.8.0
+	github.com/cometbft/cometbft v0.38.2
+	github.com/cometbft/cometbft-db v0.9.1 // indirect
 	github.com/cosmos/cosmos-proto v1.0.0-beta.3
-	github.com/cosmos/cosmos-sdk v0.47.4
-	github.com/cosmos/gogoproto v1.4.10
-	github.com/cosmos/ibc-go/v7 v7.3.0
-	github.com/ethereum/go-ethereum v1.11.6
+	github.com/cosmos/cosmos-sdk v0.50.2
```
aleem1314, line 13: Bump SDK and other dependencies to new minor version.

app/upgrades.go
```diff
@@ -266,4 +224,7 @@ func (app *App) RegisterUpgradeHandlers(semverVersion string) {
 		// configure store loader that checks if version == upgradeHeight and applies store upgrades
 		app.SetStoreLoader(upgradetypes.UpgradeStoreLoader(upgradeInfo.Height, &storeUpgrades))
 	}
+	app.UpgradeKeeper.SetUpgradeHandler("v0.50-upgrade", func(ctx context.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
```
aleem1314, line 227:
```suggestion
	app.UpgradeKeeper.SetUpgradeHandler(semverVersion, func(ctx context.Context, plan upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
```

aleem1314 (approved):
pre approving

app/upgrades.go
vishal-kanna: Okay

go.mod
vishal-kanna: Okay

taariq:
@aleem1314 we have just TWO unresolved convos on the PR. We will merge on your check and resolve.

faddat:
Just wanted to say that I thought this is cool :-)

taariq:
> Just wanted to say that I thought this is cool :-)

That means so much to use @faddat! Thank you!

PR #1071, opened 2024-01-08
WIP

byte-bandit:
# Related Github tickets

# Background

_A short paragraph about what exactly this PR solves and what is the problem._

# Testing completed

# Breaking changes


x/evm/types/eth_txable.go
```diff
@@ -4,6 +4,8 @@ import (
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 )
 
+// TODO: Implement TX verifications
```
byte-bandit, line 7: None of the message types currently evaluate TX contents as part of their execution verification. Implementing that is outside of the scope of this work, but I believe it was intended to be included at one point to tighten security.

taariq: Yeup. That's true. I believe there's still an open ticket for this.

taariq (approved):
LGTM!

Issue #1079, opened 2024-01-29
BUG: Paloma is using old version of zondax hid and ledger-go causing issues on Macos ledger nano S: hidapi: failed to open device


taariq:
# What is happening?
<details>
  <summary>Section description</summary>
<i>Provide as much context as you can. Give as much context as you can to make it easier for the developers to figure what is happening.</i>
</details>

BUG: Paloma is using old version of zondax hid and ledger-go causing issues on Macos ledger nano S: hidapi: failed to open device. It appears that we are on older version of ledger in go.mod. 

`github.com/zondax/hid v0.9.1 // indirect`
needs to be upgraded to https://github.com/Zondax/hid/releases/tag/v0.9.2
Also we may want to upgrade ledger-go
https://github.com/Zondax/ledger-go/releases/tag/v0.14.3

# Paloma and pigeon versions and logs
<details>
  <summary>Section description</summary>
<i>Write down paloma version.</i>
<i>Write down pigeon version.</i>
<i>Copy and paste pigeon config file as well as relevant ENV variables.</i>
</details>

Paloma 1.11.0 
Pigeon 1.10.2

# How to reproduce?
<details>
  <summary>Section description</summary>
  <i>Please write detailed steps of what you were doing for this bug to appear.</i>
</details>

Attempting to sign transactions with the build version of Paloma fails.

See other issues called here: https://github.com/Agoric/agoric-sdk/issues/8783

# What is the expected behaviour?
<details>
  <summary>Section description</summary>
<i>If you know, please write down what is the expected behaviour. If you don't know, that's ok. We can have a discussion in comments.</i>
</details>

taariq:
@byte-bandit this error still persists. Unable to use any command with Paloma v1.12.0 and Ledger build.

Each transaction fails to trigger the ledger device and there's no output with the error. See example below.

```
body:
  extension_options: []
  memo: ""
  messages:
  - '@type': /cosmos.bank.v1beta1.MsgSend
    amount:
    - amount: "1000000"
      denom: ugrain
    from_address: paloma18xrvj2ffxygkmtqwf3tr6fjqk3w0dgg7m6ucwx
    to_address: paloma1qxl8xy0rjl99zhaxgu4ffuegup9mh6qxzt0jqm
  non_critical_extension_options: []
  timeout_height: "0"
signatures: []
confirm transaction before signing and broadcasting [y/N]: y
Error:
Usage:
[...]

PR #1080, opened 2024-01-29
chore: upgrade zondax ledger libraries

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1233
- https://github.com/palomachain/paloma/issues/1079


taariq (approved):
Small change. big impact. LFG.

taariq:
@byte-bandit Even after the update, I am still unable to send with ledger. I think I will need some further debugging here.

Issue #1088, opened 2024-03-10
BUG: Report from community member that the chain appears to be struggling. Validators are missing blocks


taariq:
# What is happening?
<details>
  <summary>Section description</summary>
<i>Provide as much context as you can. Give as much context as you can to make it easier for the developers to figure what is happening.</i>
</details>

![image_2024-03-10_11-53-34](https://github.com/palomachain/paloma/assets/701864/b5faa00f-8815-47da-a555-a8bbfc6e4694)

ppnv:
4:01PM INF trying to deploy smart contract to EVM chain chain-reference-id=bnb-main module=x/evm smart-contract-id=6
4:01PM INF trying to deploy smart contract to EVM chain chain-reference-id=eth-main module=x/evm smart-contract-id=6
4:01PM INF trying to deploy smart contract to EVM chain chain-reference-id=gnosis-main module=x/evm smart-contract-id=6
4:01PM INF trying to deploy smart contract to EVM chain chain-reference-id=matic-main module=x/evm smart-contract-id=6
4:01PM INF trying to deploy smart contract to EVM chain chain-reference-id=op-main module=x/evm smart-contract-id=6
4:01PM INF trying to deploy smart contract to all chains module=x/evm smart-contract-id=6
4:01PM ERR no validator found for cons pub address. component=metrix.UpdateUptime module=x/metrix signed-blocks-window=10000 validator-conspub=palomavalcons1qsys7shahzl3fplvu0gzfz9tjqxqzz3e0vulsg
4:01PM ERR no validator found for cons pub address. component=metrix.UpdateUptime is-jailed=true missed-blocks-counter=32 module=x/metrix signed-blocks-window=10000 uptime=0.000000000000000000 validator=palomavaloper1d3v3jh6l2r23y9kgzdrahx0ev8ez0g8qj384sr validator-conspub=palomavalcons1pyjrnzmdh28ztnvk2fhhedstasfftcrk48cc46
4:01PM ERR no validator found for cons pub address. component=metrix.UpdateUptime is-jailed=true missed-blocks-counter=2 module=x/metrix signed-blocks-window=10000 uptime=0.000000000000000000 validator=palomavaloper10up3f4uxtsfyjkvdypag2q7vr7yxahzd7tex99 validator-conspub=palomavalcons1zpa3rkd0nr72qvwdlvv63tdddj6qpag95rgfnf
[...]

PR #1089, opened 2024-03-11
chore: fixes in the upgrade code 

vishal-kanna:
closes: #1088


byte-bandit (approved):
Nice find @vishal-kanna !

Unfortunately, I don't think this is connected to https://github.com/palomachain/paloma/issues/1088

The typo is not actually in production: https://github.com/palomachain/paloma/blob/53de2d0fe43950f197fda42e4813b28861f69ed1/x/consensus/module.go#L164

I assume it was merged to `master` with the big Cosmos SDK change and we just didn't catch it.

vishal-kanna:
Yeah it was merged with SDK change.We got the same issue ` consensus not achieved error` while running e2e tesing .So i guess this is the one which fixes this as well

vishal-kanna:
@byte-bandit @taariq Please merge this no more changes needed from our side.

Just a reminder, Please change the old upgrade handler name before whenever you are planning for a release.

PR #1090, opened 2024-03-11
feat: attestation enforcement

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1351
- https://github.com/VolumeFi/paloma/issues/1352

# Background

We're observing an increasing amount of missing attestations on message relay evidence. Due to the equalised distribution of shares across our validator set, this can lead to successful relays and other operative messages not reaching consensus within Paloma.

This change introduces a decent amount of logging around the consensus finding within message attestation. Additionally, any outstanding attestations from expected Validators of the active valset will lead to punishment by jailing. In addition to the new logging output, affected validators are incentivised to examine their setups and address any ongoing issues. 

# Testing completed

# Breaking changes


taariq (approved):
LGTM!

PR #1095, opened 2024-03-21
hotfix: prevent punishing missing attestants for unhandled msg

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1409

# Background

Some messages might still fail to be handled at all, meaning the message will not even enter the attestation phase. In this case, we do not want to punish validators - otherwise we'd jail the entire network.

Merging this into `devel`, which was taken from `release/v1.12.2`. That will be my working branch for a `v1.12.3` release in case `master` remains blocked due to the Cosmos upgrade.

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/keeper/concensus_keeper.go
+++ b/x/consensus/keeper/concensus_keeper.go
@@ -281,6 +281,12 @@ func (k Keeper) jailValidatorsWhichMissedAttestation(ctx sdk.Context, queueTypeN
 		return fmt.Errorf("getMsgByID: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


taariq:
@byte-bandit minor release will require chain halt. Yes?

byte-bandit:
> @byte-bandit minor release will require chain halt. Yes?

It touches consensus operations, so yes. This will need a full governance update. I will try to squeeze in the next fix for the stuck gravity batch as well.

PR #1099, opened 2024-03-28
fix: prevent TPK when not attesting faulty msg

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1425

# Background

This change adds two safe guards to prevent a mass jailing in case of a faulty message:

a) messages that do not build consensus because neither `errorData` nor `publicAccessData` has been added, meaning the message was never handled in the first place. Nobody should be punished for not attesting what's not attestable.
b) messages that less than 10% of the network attest are treated as indications of faulty message responses which cannot be attested to even when tried. Nobody will be punished for missing attestation on such a message.

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/keeper/cleanup_test.go
+++ b/x/consensus/keeper/cleanup_test.go
@@ -6,9 +6,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/palomachain/paloma/x/consensus/keeper/consensus"
 	"github.com/palomachain/paloma/x/consensus/types"
-	valsettypes "github.com/palomachain/paloma/x/valset/types"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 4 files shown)


x/consensus/keeper/cleanup_test.go
```diff
@@ -64,7 +62,7 @@ func TestDeleteOldMessages(t *testing.T) {
 
 				// Advance to block 61
 				ctx = ctx.WithBlockHeight(61)
-				ms.ValsetKeeper.On("GetCurrentSnapshot", mock.Anything).Return(&valsettypes.Snapshot{}, nil)
+				// ms.ValsetKeeper.On("GetCurrentSnapshot", mock.Anything).Return(&valsettypes.Snapshot{}, nil)
```
byte-bandit, line 65:
```suggestion
```

taariq (approved):
LGTM!

PR #1100, opened 2024-03-29
chore: update README and min pigeon version

byte-bandit:


README.md
```diff
@@ -88,7 +88,7 @@ See [Release procedure](CONTRIBUTING.md#release-procedure) for more information
 The current required version of libwasm is `1.3.0`. If you're upgrading from a prior version it is recommended to remove the cache to avoid errors. If you're already have `palomad` running, you will need to stop it before doing these steps.
```
verabehr, line 88: should be updated as well here. The current required version of libwasm is `1.5.2`

verabehr (changes requested):

README.md
byte-bandit, line 88:
```suggestion
The current required version of libwasm is `1.5.2`. If you're upgrading from a prior version it is recommended to remove the cache to avoid errors. If you're already have `palomad` running, you will need to stop it before doing these steps.
```

byte-bandit: Good catch. Done!

verabehr (approved):

PR #1116, opened 2024-04-08
chore: removed the old upgrade handler

vishal-kanna:


byte-bandit:
I'm not sure the old upgrade handler is responsible here, since it's just a set of conditionals that would not be executed. But it's worth a try. We'd lose the ability to restore from an old snapshot, but I think that's no problem @taariq @verabehr ?

@vishal-kanna Did you test this in pvt test net? Setting it up with `v1.13.0`, proposing a software upgrade and successfully upgrading to a version built with these changes?

vishal-kanna:
@byte-bandit I've tested it locally.It is working fine and i need to test it in the pvt testnet.

byte-bandit:
@vishal-kanna Sounds good. Since this change addresses the upgrade mechanic, we need to ensure that a network will be able to upgrade successfully from `v1.13.0` to `v1.13.1`.

taariq:
@vishal-kanna when will you have a test completed on this? We would like to cut a new release with this fix and the upgrade on CometBFT.

byte-bandit (approved):

PR #1132, opened 2024-04-16
chore: increase go version to 1.22.2 and fix libwasm version

byte-bandit:
- Increase Go version to `1.22.2`
- Fix libwasm version during application startup


taariq (approved):
LFG!

PR #1135, opened 2024-04-17
Add testnet command

alpe:
Adds command to setup a testnet:

```sh
palomad testnet init-files --chain-id=testing --output-dir=./testnet --v=4 --keyring-backend=test --minimum-gas-prices=0.000001ugrain --staking-denom=ugrain --commit-timeout=900ms --single-host --starting-ip-address=""
```

Disclaimer: 
copied from wasmd which came from cosmos-sdk/simapp with some modifications

### DIFF
--- a/cmd/palomad/commands.go
+++ b/cmd/palomad/commands.go
@@ -7,6 +7,7 @@ import (
 	cosmoslog "cosmossdk.io/log"
 	confixcmd "cosmossdk.io/tools/confix/cmd"
 	"github.com/CosmWasm/wasmd/x/wasm"
+	wasmcli "github.com/CosmWasm/wasmd/x/wasm/client/cli"
 	dbm "github.com/cosmos/cosmos-db"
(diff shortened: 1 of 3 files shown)


cmd/palomad/commands.go
```diff
@@ -111,6 +114,7 @@ func initRootCmd(
 		txCommand(basicManager),
 		keys.Commands(),
 	)
+	wasmcli.ExtendUnsafeResetAllCmd(rootCmd)
```
alpe, line 117: This clears the `wasm` folder on unsafe-reset-all, too. Not related to the testnet command though

taariq: @byte-bandit do we want to include this in paloma?

taariq: Yup This is it.

taariq:
@alpe quick lint check?

taariq (approved):

PR #1139, opened 2024-04-18
chore: updated the iavl deps

vishal-kanna:


byte-bandit (changes requested):
Blocking this for now until we have `v1.13.2` out the door. There's a lot of overrides in here, I think it's more sensible to wait until the IAVL updates are ported into a regular Cosmos upgrade if we don't need them.

PR #1140, opened 2024-04-18
fix: include v prefix in app version

byte-bandit:
The leading v prefix in the app version is used both in CI builds
as well as the test net setup. Pigeon also uses it. It makes
no sense to remove it from manual builds, especially as it can
lead to differences in state.


taariq (approved):
LFG!

PR #1166, opened 2024-05-20
chore: add codeql action and config

maharifu:
Add codeql github action and config based on cosmos-sdk config - https://github.com/cosmos/cosmos-sdk/blob/main/.github/workflows/codeql-analysis.yml

# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1476

# Background

Add a new github action to run CodeQL on the codebase. It add no functionality, so there are no tests and there should not be any breaking changes.

# Testing completed

# Breaking changes


byte-bandit (approved):
LGTM, it's already picked up quite a handful: https://github.com/palomachain/paloma/security/code-scanning?query=pr%3A1166+is%3Aopen

@verabehr I think it's worth creating a follow-up ticket to address the issues reported by this, especially around the panics, as these can lead to the infamous AppHash.

verabehr:
ticket added here https://github.com/VolumeFi/paloma/issues/1591

PR #1168, opened 2024-05-23
Add feature set to weight proposal CLI command

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1210

# Background

The `propose-relay-weights` CLI command was missing the `featureSet` weight. This PR gives users the ability to also define this weight when submitting a proposal.

# Testing completed

# Breaking changes


byte-bandit (approved):
Nice work, lgtm!

PR #1175, opened 2024-06-03
Periodically update xchain reference blocks

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1637

# Background

At startup, pigeon validates that it's talking to the right chain by using the reference block height and hash. These values are introduced when first integrating a chain. However, as time passes, some RPC providers will prune block information, which makes it impossible for pigeon to verify the blocks.

To solve this, we introduce a new periodic task that at each 10000 blocks, sends a message requesting pigeons to update the reference block information (height and hash) for each supported chain.

We add a new queue - `evm/<CHAIN>/reference-block` - and follow the same flow from ValidatorBalances.

Related pigeons changes are handled in https://github.com/palomachain/pigeon/pull/393

# Testing completed

# Breaking changes


x/evm/keeper/attest_reference_block.go
```diff
@@ -0,0 +1,66 @@
...
+	logger.Debug("attest-reference-block")
+
+	cacheCtx, writeCache := sdkCtx.CacheContext()
+	defer func() {
+		if retErr == nil {
+			writeCache()
+		}
+	}()
+
+	result, err := k.consensusChecker.VerifyEvidence(cacheCtx, msg.GetEvidence())
```
byte-bandit, line 35: Can we extract this? We have this 3 times in our code base now.

x/evm/keeper/attest_reference_block.go
```diff
@@ -0,0 +1,66 @@
...
+			return nil
+		}
+		return err
+	}
+
+	defer func() {
+		// given that there was enough evidence for a proof, regardless of the outcome,
+		// we should remove this from the queue as there isn't much that we can do about it.
+		if err := q.Remove(cacheCtx, msg.GetId()); err != nil {
+			k.Logger(sdkCtx).Error("error removing message, attestReferenceBlock", "msg-id", msg.GetId(), "msg-nonce", msg.Nonce())
```
byte-bandit, line 52: This as well. Can we maybe extract this kind of "house keeping" into something like a wrapper or orchestrator that governs all these attestations?

x/evm/keeper/attest_reference_block.go
```diff
@@ -0,0 +1,66 @@
...
+
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	logger := k.Logger(ctx).WithFields(
+		"chain-id", chainReferenceID,
+		"component", "attest-reference-block",
+		"msg-id", msg.GetId(),
+		"msg-nonce", msg.Nonce())
+	logger.Debug("attest-reference-block")
+
+	cacheCtx, writeCache := sdkCtx.CacheContext()
```
byte-bandit, line 28: I don't think this is needed here tbh. There's only 1 state change involved here, which we will only reach if all preconditions are fulfilled.

x/evm/keeper/keeper.go
```diff
@@ -169,6 +178,16 @@ func (k Keeper) ChangeMinOnChainBalance(ctx sdk.Context, chainReferenceID string
 	return k.updateChainInfo(ctx, ci)
 }
 
+func (k Keeper) ChangeChainReferenceBlock(ctx sdk.Context, chainReferenceID string, referenceBlockHeight uint64, referenceBlockHash string) error {
```
byte-bandit, line 181: Just nickpicking

```suggestion
func (k Keeper) UpdateChainReferenceBlock(ctx sdk.Context, chainReferenceID string, referenceBlockHeight uint64, referenceBlockHash string) error {
```

x/evm/keeper/keeper.go
```diff
@@ -169,6 +178,16 @@ func (k Keeper) ChangeMinOnChainBalance(ctx sdk.Context, chainReferenceID string
...
+	ci, err := k.GetChainInfo(ctx, chainReferenceID)
+	if err != nil {
+		return err
+	}
+	ci.ReferenceBlockHeight = referenceBlockHeight
```
byte-bandit, line 186: Let's make sure the new height is larger than the last one.

x/evm/keeper/keeper.go
```diff
@@ -680,6 +699,25 @@ func (k Keeper) CheckExternalBalancesForChain(ctx context.Context, chainReferenc
 	return err
 }
 
+func (k Keeper) UpdateReferenceBlockForChain(ctx context.Context, chainReferenceID string) error {
```
byte-bandit, line 702: More nitpicking

```suggestion
func (k Keeper) ScheduleReferenceBlockForChain(ctx context.Context, chainReferenceID string) error {
```

x/evm/module.go
```diff
@@ -192,5 +194,23 @@ func (am AppModule) EndBlock(ctx context.Context) error {
...
+				return
+			}
+
+			for _, ci := range cis {
+				err := am.keeper.UpdateReferenceBlockForChain(sdkCtx, ci.GetChainReferenceID())
+				if err != nil {
+					am.keeper.Logger(sdkCtx).Error("error while adding request to refresh reference block for chain", "error", err, "chain-reference-id", ci.GetChainReferenceID())
+				}
+			}
+		}()
```
byte-bandit, line 212: Why not extract this into testable function that returns an error. That way you can just call it here, and log once.

Also, use `liblog` for logging.

byte-bandit (changes requested):
Solid solution. Got some small remarks.

This also isn't in Genesis yet, but I guess it doesn't really have to be.

x/evm/keeper/attest_reference_block.go
maharifu: I moved this and the previous into a wrapper and refactored the relevant functions.

x/evm/keeper/attest_reference_block.go
maharifu: I kept it as part of the wrapper as other handler might need it.

byte-bandit (commented):
Needs a rebase, the query server stuff is on master already. Also a FIXME? 

Other than that it looks good!

x/evm/keeper/attest.go
```diff
@@ -21,33 +21,27 @@ import (
...
+type msgAttester func(sdk.Context, consensus.Queuer, consensustypes.QueuedSignedMessageI, any) error
 
+func (k Keeper) attestMessageWrapper(ctx context.Context, q consensus.Queuer, msg consensustypes.QueuedSignedMessageI, fn msgAttester) (retErr error) {
 	if len(msg.GetEvidence()) == 0 {
 		return nil
 	}
 
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	logger := k.Logger(ctx).WithFields(
+		"component", "attest-reference-block", // FIXME wrong component
```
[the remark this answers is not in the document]
maharifu: :facepalm: This wasn't supposed to make it to the PR, but I forgot to change it!

byte-bandit (approved):
🚀

PR #1180, opened 2024-06-05
Move minimum required pigeon version from hard-coded to governance

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1640

# Background

At the moment, we store the minimum required version of Pigeon as hard-coded const inside Paloma. This means to enforce a Pigeon update, we need to roll-out a new version of the protocol as well, which often creates unnecessary noise and maintenance overhead.

Instead, this information should live on the chain directly, and be addressable via governance vote.

- Generalize the restriction to pigeon requirements, so we can add new restriction in the future, instead of only minimum version
- Add pigeon requirements to valset keeper, as the valset module is the one requiring this information
- Add scheduled pigeon requirements, so we can schedule rollouts to a specific block height
- Add new gRPC query - `q valset get-pigeon-requirements` - to get current and scheduled requirements
- Add new tx - `tx gov submit-legacy-proposal valset propose-pigeon-requirements` - to propose changes to pigeon requirements
- The valset module checks for pending scheduled changes at each `BeginBlock`. When a change is applied, scheduled changes are cleared
- Added this information to genesis

# Testing completed

# Breaking changes


x/valset/keeper/keeper.go
```diff
@@ -25,12 +25,22 @@ import (
...
 	cJailingNetworkShareProtection  = 0.25
+	// We set the default to the last pigeon version enforced in code to keep
+	// backwards compatibility
+	defaultMinimumPigeonVersion = "v1.11.2"
+)
+
+var (
+	pigeonStoreKey                 = []byte("pigeon-requirements")
+	pigeonRequirementsKey          = []byte("current")
+	pigeonScheduledRequirementsKey = []byte("scheduled")
```
byte-bandit, line 43: Keys usually live in the `types` package in `key.go` or so.

x/valset/keeper/keeper.go
```diff
@@ -729,3 +736,96 @@ func (k Keeper) SaveModifiedSnapshot(ctx context.Context, snapshot *types.Snapsh
...
+		return errors.New("new version cannot be lower than current")
+	}
+
+	err = keeperutil.Save(pigeonStore, k.cdc, pigeonRequirementsKey, req)
+	if err != nil {
+		return err
+	}
+
+	// Delete scheduled requirements
+	pigeonStore.Delete(pigeonScheduledRequirementsKey)
```
byte-bandit, line 769: What happens in case the save goes through but the delete fails? Will it be re-attempted and then continously fail since the version isn't higher anymore?

If so, use a branched context and only commit in case call operations go through.

byte-bandit (commented):
Good stuff. 2 small remarks

x/valset/keeper/keeper.go
maharifu: I added the CacheContext to both, but I'm not 100% sure it's right. The `Delete` method does not return an error, so it might still be possible to fall into the same situation?

We could be more defensive about this by having `SetPigeonRequirements` return a specific error for version mismatch. This could be detected in `BeginBlock` and we clear the value explicitly from there. What do you think?

byte-bandit: Hmm. Maybe the more straight-forward solution would be to alter the `Delete` method to return an error. I wonder that's not happening. Is it coming straight from the Cosmos framework or is there some wrapper magic happening around it?

maharifu: It is coming straight from the SDK. We get a `KVStore` that embeds a [BasicKVStore](https://github.com/cosmos/cosmos-sdk/blob/main/store/types/store.go#L233) that does not return errors on any operation. The `Save` is [our own wrapper](https://github.com/palomachain/paloma/blob/master/util/keeper/save_load.go#L10) and just returns marshalling errors.

maharifu: It seems the concrete store implementations [may panic](https://github.com/cosmos/cosmos-sdk/blob/main/store/iavl/store.go#L198) instead of returning an error. In that case, the CacheContext and commit after Delete should work as all-or-nothing and should be fine. There may still be the case where there's some fs corruption affecting one store and not the other, but the node will have bigger problems then.

PR #1183, opened 2024-06-05
fix: gravity genesis export

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1654

# Background

The `GravityNonces` changed to an array, so now we have to call `SetLastSlashedBatchBlock` manually if it's empty.

# Testing completed

# Breaking changes


byte-bandit (approved):
🚀

PR #1188, opened 2024-06-07
Enable compass deployment retries

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1035

# Background

As a user, I want failed compass deployments to be retried in a sane way in case they failed, so that I'm not locked out of an upgraded system and my funds are protected.

- Added a `Retries` field to `UploadSmartContract`
- If deployment fails and we haven't reached max retries, we push the message back into the queue
- When max retries is reached, we remove the message from the queue

Also, re-enabled and fixed ginkgo tests in evm keeper.

# Testing completed

# Breaking changes


byte-bandit (approved):
Good stuff!

PR #1195, opened 2024-06-14
fix: withold gravity batches until valset update

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1464

# Background

This change adds a new constraint when handing out gravity batches to be relayed. They will now be withheld while valset updates are pending on the target chain, just like with SLC. I had to cut some corners to get this out sooner rather than later, but I also added a more decentralised event bus system for decoupling systems. It's currently very bare bone and as singleton hard to test, but if it grows, we can refactor into something more injectible.

# Testing completed

# Breaking changes

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -595,6 +595,7 @@ func New(
 		app.DistrKeeper,
 		app.TransferKeeper,
 		app.EvmKeeper,
+		app.ConsensusKeeper,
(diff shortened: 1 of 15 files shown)


maharifu (approved):
LGTM

PR #1197, opened 2024-06-20
feat: add relayer fee settings

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1710

# Background

> This change adds the concept of individual
> relayer fees that may be set by any active
> validator for any supported target chain.

# Testing completed

# Breaking changes


x/treasury/client/cli/tx.go
```diff
@@ -21,5 +25,69 @@ func GetTxCmd() *cobra.Command {
...
+	return cmd
+}
+
+func CmdUpsertRelayerFee() *cobra.Command {
+
+	cmd := &cobra.Command{
+		Use:   "upsert-relayer-fee [chain-reference-id] [fee-multiplicator]",
+		Short: "Sets the relayer fee for the sender to the given multiplicator.",
+		Long: `
+    Sets the relayer fee for the sender to th given multiplicator value.
```
maharifu, line 38:
```suggestion
    Sets the relayer fee for the sender to the given multiplicator value.
```

x/treasury/client/cli/tx.go
```diff
@@ -21,5 +25,69 @@ func GetTxCmd() *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+
+			id := args[0]
+			fee := math.LegacyMustNewDecFromStr(args[1])
+
+			if fee.LT(math.LegacyZeroDec()) {
+				return fmt.Errorf("negative fee not allowed")
+			}
```
maharifu, line 64: Is this needed? We're not allowing values less than 1 below.

maharifu (changes requested):
Nice one! Just a couple of nitpicks.

proto/palomachain/paloma/treasury/query.proto
```diff
@@ -19,6 +21,11 @@ service Query {
   rpc QueryFees(QueryFeesRequest) returns (Fees) {
     option (google.api.http).get = "/palomachain/paloma/treasury/fees";
   }
+
+  // RelayerFee queries the fee settings of a validator participating in consensus.
+  rpc RelayerFee(QueryRelayerFeeRequest) returns (RelayerFeeSetting) {
+    option (google.api.http).get = "/palomachain/paloma/treasury/relayer-fee";
```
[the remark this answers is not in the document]
byte-bandit: I don't know. While there's no clear answer, general consensus seems to shift towards using hyphens over underscores. It's mostly for SEO, but I've become more used to them now.

That being said, all other URIs are using underscore, so better to stick with one than mix.

maharifu (approved):
LGTM

PR #1198, opened 2024-06-20
feat: daily bridge transfer limits

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1711

# Background

Add a limit to the amount of tokens transferred out from paloma within a specified window, per token. This is defined by a new governance vote.

- Add governance vote. Can define token, window size, token limit and list of exempt addresses.
- Keep a counter of the number of tokens transferred and the block height at the start of the limiting window.
- Before accepting a transfer, update the counters and check the limits.
- Transfers are refused if over the transfer limit.

# Testing completed

# Breaking changes


maharifu:
@taariq @byte-bandit I assumed we want to update/check the counters when adding the transfer to the queue, and will not decrease the counters even if the transfer fails or is cancelled afterwards.

taariq:
@maharifu correct. No need to decrease the counters on cancellation. All good.

taariq:
@byte-bandit your review? Let's merge this in!

proto/palomachain/paloma/gravity/query.proto
```diff
@@ -48,6 +49,12 @@ service Query {
   rpc GetBridgeTax(google.protobuf.Empty) returns (QueryBridgeTaxResponse) {
     option (google.api.http).get = "/palomachain/paloma/gravity/bridge_tax";
   }
+
+  rpc GetAllBridgeTransferLimits(google.protobuf.Empty)
```
byte-bandit, line 53:
```suggestion
  rpc GetBridgeTransferLimits(google.protobuf.Empty)
```

x/gravity/client/cli/query.go
```diff
@@ -366,3 +367,33 @@ func GetCmdQueryBridgeTax() *cobra.Command {
 
 	return cmd
 }
+
+// GetCmdQueryAllBridgeTransferLimits fetches transfer limits for all tokens
+func GetCmdQueryAllBridgeTransferLimits() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "all-bridge-transfer-limits",
```
byte-bandit, line 374:
```suggestion
		Use:   "bridge-transfer-limits",
```

x/gravity/keeper/grpc_query_get_all_bridge_transfer_limits.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	keeperutil "github.com/palomachain/paloma/util/keeper"
+	"github.com/palomachain/paloma/x/gravity/types"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+func (k Keeper) GetAllBridgeTransferLimits(
```
byte-bandit, line 13:
```suggestion
func (k Keeper) GetBridgeTransferLimits(
```

byte-bandit, line 13: FYI, I've been thinking a bit about API design on the keepers lately. The most convincing argument I've come across is to 

- start functions related to gRPC queries with `Query...`
- start functions on message server implementation with the verb, i.e. `Relay...`, `Start...`, etc...
- start simple getter functions on the keeper implementation with `Get`
- name more advanced getters by their index, i.e. `Get...ByID`
- start iteration callbacks with `Iterate...`

It's not perfect. Personally, I'd prefer calling getters by their entity, and setters by their function, i.e. `BridgeLimit`, `BridgeLimits`, `BridgeLimitByID`, `UpdateBridgeLimit`, etc...

But the thoughts outlined above seem to be more consistent with the cosmos SDK coding style. Just some food for thought here.

byte-bandit (dismissed):
Good stuff. Some small input, but non-blocking. Up to you.

proto/palomachain/paloma/gravity/bridge_transfer_limit_proposal.proto
```diff
@@ -0,0 +1,18 @@
...
+
+message SetBridgeTransferLimitProposal {
+  string title = 1;
+  string description = 2;
+  string token = 3;
+  string limit = 4 [
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false
+  ];
+  string limit_period = 5;
```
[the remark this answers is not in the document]
maharifu: Nice catch. I was converting it in the keeper, but it makes sense to convert it earlier in the command and return an error then. :+1:

byte-bandit (approved):

PR #1199, opened 2024-06-21
fix: unpack errors on ReferenceBlockAttestation

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1740

# Background

ReferenceBlockAttestation messages currently do not need an assignee, however they are still queried for messages to relay, leading to unpack error messages in palomad logs.
By first checking for publicAccessData, ReferenceBlockAttestation messages are filtered before we try to unpack them as TurnstoneMsg and avoid the errors.

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/keeper/concensus_keeper.go
+++ b/x/consensus/keeper/concensus_keeper.go
@@ -183,6 +183,11 @@ func (k Keeper) GetMessagesForRelaying(ctx context.Context, queueTypeName string
 		return msg.GetId() <= valsetUpdatesOnChain[0].GetId()
 	})
 
+	// Filter down to just messages that have neither publicAccessData nor errorData
+	msgs = slice.Filter(msgs, func(msg types.QueuedSignedMessageI) bool {
+		return msg.GetPublicAccessData() == nil && msg.GetErrorData() == nil
+	})
+
 	// Filter down to just messages assigned to this validator
 	msgs = slice.Filter(msgs, func(msg types.QueuedSignedMessageI) bool {
 		var unpackedMsg evmtypes.TurnstoneMsg
@@ -192,11 +197,6 @@ func (k Keeper) GetMessagesForRelaying(ctx context.Context, queueTypeName string
 		}
 
 		return unpackedMsg.GetAssignee() == valAddress.String()
-	})
-
-	// Filter down to just messages that have neither publicAccessData nor errorData
-	msgs = slice.Filter(msgs, func(msg types.QueuedSignedMessageI) bool {
-		return msg.GetPublicAccessData() == nil && msg.GetErrorData() == nil
 	})
 
 	if len(msgs) > defaultResponseMessageCount {


byte-bandit (changes requested):
Take a look at `ValidatorBalanceAttestation` messages. They have the same constraints (no relaying), but the logic is handled differently (by populating some `PublicAccessData` I believe).

How about using the same approach here, so we don't end up with `n` solutions for the same problem again?

maharifu:
I think `ValidatorBalancesAttestation` works by having `Assignee` defined [in the proto](https://github.com/palomachain/paloma/blob/master/proto/palomachain/paloma/evm/evm_state_attestation.proto#L17) although it doesn't need it as well. This makes it comply with the `TurnstoneMsg` interface.

I didn't want to add the field to `ReferenceBlockAttestation` as it's not needed. We can remove it from `ValidatorBalancesAttestation` and implement an empty function as well, but didn't want to change it unnecessarily. Do you think we should remove it?

byte-bandit:
> I think `ValidatorBalancesAttestation` works by having `Assignee` defined [in the proto](https://github.com/palomachain/paloma/blob/master/proto/palomachain/paloma/evm/evm_state_attestation.proto#L17) although it doesn't need it as well. This makes it comply with the `TurnstoneMsg` interface.
> 
> I didn't want to add the field to `ReferenceBlockAttestation` as it's not needed. We can remove it from `ValidatorBalancesAttestation` and implement an empty function as well, but didn't want to change it unnecessarily. Do you think we should remove it?
> 
> **edit**
> 
> We set `PublicAccessData` in [ReferenceBlockAttestation messages](https://github.com/palomachain/paloma/blob/master/x/evm/keeper/keeper.go#L731), but pigeon still queries all queues for relay [here](https://github.com/palomachain/pigeon/blob/master/relayer/message_relayer.go#L66). I think we could also address this in pigeon by splitting [SupportedQueues](https://github.com/palomachain/pigeon/blob/master/chain/evm/processor.go#L36) but that also doesn't feel like addressing the root of the problem.

Pigeon will still query it, but Paloma won't hand the message out for delivery, since the `PublicAccessData` is already set. See `GetMessagesForRelaying` in the consensus keeper. Essentially, setting the `PublicAccessData` field to anything but `nil` should prevent this message from ever being handed over to the `assignee`.  It will also trigger the "attestation" stage of the message lifecycle, since Paloma will start handing it over to all Pigeons for attestation once either `PublicAccessData` or `ErrorData` are filled.

To be honest, I think it's more in line with the current architecture to stick with this approach. Can you see any reasons why you'd want to move away from it?

Splitting queues is likely something we want to explore in the future, but out of scope for this one.

maharifu:
> > I think `ValidatorBalancesAttestation` works by having `Assignee` defined [in the proto](https://github.com/palomachain/paloma/blob/master/proto/palomachain/paloma/evm/evm_state_attestation.proto#L17) although it doesn't need it as well. This makes it comply with the `TurnstoneMsg` interface.
> > I didn't want to add the field to `ReferenceBlockAttestation` as it's not needed. We can remove it from `ValidatorBalancesAttestation` and implement an empty function as well, but didn't want to change it unnecessarily. Do you think we should remove it?
> > **edit**
> > We set `PublicAccessData` in [ReferenceBlockAttestation messages](https://github.com/palomachain/paloma/blob/master/x/evm/keeper/keeper.go#L731), but pigeon still queries all queues for relay [here](https://github.com/palomachain/pigeon/blob/master/relayer/message_relayer.go#L66). I think we could also address this in pigeon by splitting [SupportedQueues](https://github.com/palomachain/pigeon/blob/master/chain/evm/processor.go#L36) but that also doesn't feel like addressing the root of the problem.
> 
> Pigeon will still query it, but Paloma won't hand the message out for delivery, since the `PublicAccessData` is already set. See `GetMessagesForRelaying` in the consensus keeper. Essentially, setting the `PublicAccessData` field to anything but `nil` should prevent this message from ever being handed over to the `assignee`. It will also trigger the "attestation" stage of the message lifecycle, since Paloma will start handing it over to all Pigeons for attestation once either `PublicAccessData` or `ErrorData` are filled.
> 
> To be honest, I think it's more in line with the current architecture to stick with this approach. Can you see any reasons why you'd want to move away from it?
> 
> Splitting queues is likely something we want to explore in the future, but out of scope for this one.

We have that, and it's working. I agree we should keep it.

The current issue is that in `GetMessagesForRelaying` we will try to unpack these messages as `TurnstoneMsg` [here](https://github.com/palomachain/paloma/blob/master/x/consensus/keeper/concensus_keeper.go#L189), but this will fail as we don't have the `GetAssignee()` method, as there's no `Assignee` field. Everything still works fine, but we're getting a bunch of errors in the logs, which is what I'm trying to avoid here.

I realize now a simpler solution could be to push [this check](https://github.com/palomachain/paloma/blob/master/x/consensus/keeper/concensus_keeper.go#L197) for `PublicAccessData` to happen before checking the assignee. This would filter these messages before we try to unpack them as `TurnstoneMsg` and get rid of the errors as well. Do you see an issue with doing this?

byte-bandit:
> > > I think `ValidatorBalancesAttestation` works by having `Assignee` defined [in the proto](https://github.com/palomachain/paloma/blob/master/proto/palomachain/paloma/evm/evm_state_attestation.proto#L17) although it doesn't need it as well. This makes it comply with the `TurnstoneMsg` interface.
> > > I didn't want to add the field to `ReferenceBlockAttestation` as it's not needed. We can remove it from `ValidatorBalancesAttestation` and implement an empty function as well, but didn't want to change it unnecessarily. Do you think we should remove it?
> > > **edit**
> > > We set `PublicAccessData` in [ReferenceBlockAttestation messages](https://github.com/palomachain/paloma/blob/master/x/evm/keeper/keeper.go#L731), but pigeon still queries all queues for relay [here](https://github.com/palomachain/pigeon/blob/master/relayer/message_relayer.go#L66). I think we could also address this in pigeon by splitting [SupportedQueues](https://github.com/palomachain/pigeon/blob/master/chain/evm/processor.go#L36) but that also doesn't feel like addressing the root of the problem.
> > 
> > 
> > Pigeon will still query it, but Paloma won't hand the message out for delivery, since the `PublicAccessData` is already set. See `GetMessagesForRelaying` in the consensus keeper. Essentially, setting the `PublicAccessData` field to anything but `nil` should prevent this message from ever being handed over to the `assignee`. It will also trigger the "attestation" stage of the message lifecycle, since Paloma will start handing it over to all Pigeons for attestation once either `PublicAccessData` or `ErrorData` are filled.
> > To be honest, I think it's more in line with the current architecture to stick with this approach. Can you see any reasons why you'd want to move away from it?
> > Splitting queues is likely something we want to explore in the future, but out of scope for this one.
> 
> We have that, and it's working. I agree we should keep it.
> 
> The current issue is that in `GetMessagesForRelaying` we will try to unpack these messages as `TurnstoneMsg` [here](https://github.com/palomachain/paloma/blob/master/x/consensus/keeper/concensus_keeper.go#L189), but this will fail as we don't have the `GetAssignee()` method, as there's no `Assignee` field. Everything still works fine, but we're getting a bunch of errors in the logs, which is what I'm trying to avoid here.
> 
> I realize now a simpler solution could be to push [this check](https://github.com/palomachain/paloma/blob/master/x/consensus/keeper/concensus_keeper.go#L197) for `PublicAccessData` to happen before checking the assignee. This would filter these messages before we try to unpack them as `TurnstoneMsg` and get rid of the errors as well. Do you see an issue with doing this?

I think it sounds like a great idea :)

byte-bandit (approved):
Noice

PR #1200, opened 2024-06-25
Lcarvalho/159 deploy smart contracts

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/159

# Background

Deploy any smart contract to any supported Paloma chain via the CLI using the Pigeon Feed functionality for paying for message relays.

The PR adds a CLI interface to interact with user-defined smart contracts:

- Upload new contract to Paloma. The contract will be available to be deploy in any of our target chains.
- List contracts. Users can see their contracts and deployment status, as well as target addresses when successfully deployed.
- Remove contract. Delete the contract from Paloma.
- Deploy contract. Trigger deployment of a previously uploaded smart contract to any of our target chains.

This functionality is dependent on Pigeon feed implementation on pigeon and compass. We should wait to merge this until that work is finalized.

# Testing completed

# Breaking changes


proto/palomachain/paloma/evm/query.proto
```diff
@@ -41,6 +43,13 @@ service Query {
     option (google.api.http).get =
         "/palomachain/paloma/evm/query_get_smart_contract_deployments";
   }
+
+  // Get a list of user smart contracts
+  rpc GetUserSmartContracts(QueryUserSmartContractsRequest)
```
byte-bandit, line 48: I think we should try to keep Request/Response objects in line with terminology.

```suggestion
  rpc QueryUserSmartContracts(QueryUserSmartContractsRequest)
```

proto/palomachain/paloma/evm/tx.proto
```diff
@@ -39,3 +35,45 @@ message MsgRemoveSmartContractDeploymentRequest {
...
 }
 message RemoveSmartContractDeploymentResponse {}
+
+message MsgUploadUserSmartContractRequest {
+  option (cosmos.msg.v1.signer) = "metadata";
+
+  palomachain.paloma.valset.MsgMetadata metadata = 1
+      [ (gogoproto.nullable) = false ];
+
+  string val_address = 2 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
```
byte-bandit, line 45: Probably not needed as part of the request. We can extract the value from the Metadata, right?

proto/palomachain/paloma/evm/tx.proto
```diff
@@ -39,3 +35,45 @@ message MsgRemoveSmartContractDeploymentRequest {
...
+  uint64 id = 1;
+}
+
+message MsgRemoveUserSmartContractRequest {
+  option (cosmos.msg.v1.signer) = "metadata";
+
+  palomachain.paloma.valset.MsgMetadata metadata = 1
+      [ (gogoproto.nullable) = false ];
+
+  string val_address = 2 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
```
byte-bandit, line 62: Same here.

proto/palomachain/paloma/evm/user_smart_contract.proto
```diff
@@ -0,0 +1,31 @@
...
+}
+
+// UserSmartContract defines user-uploaded smart contracts
+// We keep them in storage, so users can decide to deploy them anywhere
+message UserSmartContract {
+  message Deployment {
+    string chain_reference_id = 1;
+    DeploymentStatus status = 2;
+    string address = 3;
+  }
```
byte-bandit, line 22: Careful storing deployment data as part of the contract. I assume we're using the same architectural approach as we do with the Compass smart contract, or Jobs - meaning we store the data on chain first, then we deploy it as a second step.

Keep in mind we can deploy the contract to the same target chain more than once. We also want to consider the possibility of updating a smart contract. Probably something to consider as an upgrade path in the future.

I suggest to add `created_at` and `updated_at` block heights to this, and make sure consuming logic allows for multiple entries of same `chain_reference_id`. In the future, if we do allow upgrades, they'd have to be versioned as well.

Also, what happens if a user removes the smart contract again? All deployment history will be lost as well. I suppose that's okay, just double checking.

proto/palomachain/paloma/evm/user_smart_contract.proto
```diff
@@ -0,0 +1,31 @@
...
+
```
byte-bandit, line 23: One more thing, we probably want to make sure we evict contracts from the DB after a certain amount of block height inactivity to avoid chain bloat. 

@taariq That means we'd clear the data on Paloma after `n` blocks, so if I wanted to deploy the same contract to a new chain, I'd have to register it again. To be clear, the deployments that already happened won't be touched, they stay of course. If we leave this data on chain forever, it will grow and grow and grow. My suggestion would be to evict contracts which haven't been deployed to any chains in 30 days.

x/evm/client/cli/tx_deploy_smart_contract.go
```diff
@@ -1,40 +0,0 @@
...
-
-	"github.com/VolumeFi/whoops"
-	"github.com/cosmos/cosmos-sdk/client"
-	"github.com/cosmos/cosmos-sdk/client/tx"
-	"github.com/cosmos/cosmos-sdk/x/gov/client/cli"
-	"github.com/palomachain/paloma/x/evm/types"
-	"github.com/spf13/cobra"
-)
-
-func CmdDeploySmartContract() *cobra.Command {
```
byte-bandit, line 14: What happened to this? We still need this for compass, no?

x/evm/keeper/grpc_user_smart_contracts.go
```diff
@@ -0,0 +1,27 @@
...
+	req *types.QueryUserSmartContractsRequest,
+) (*types.QueryUserSmartContractsResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	contracts, err := k.UserSmartContracts(ctx, req.ValAddress)
+	if err != nil {
+		return nil, err
+	}
```
byte-bandit, line 22: You might want to catch and handle `keeperutil.ErrNotFound`. I always find the message somewhat too cryptic to return to the CLI. But that's just me.

x/evm/client/cli/tx_deploy_smart_contract.go
byte-bandit: Ah, of course it was the dummy message, not the one coming from the proposal.

x/evm/keeper/msg_server_user_smart_contract.go
```diff
@@ -0,0 +1,57 @@
...
+	ctx context.Context,
+	req *types.MsgUploadUserSmartContractRequest,
+) (*types.MsgUploadUserSmartContractResponse, error) {
+	contract := &types.UserSmartContract{
+		ValAddress:       req.ValAddress,
+		Title:            req.Title,
+		AbiJson:          req.AbiJson,
+		Bytecode:         req.Bytecode,
+		ConstructorInput: req.ConstructorInput,
+	}
```
byte-bandit, line 20: I suggest performing a bit of arbitrary input validation here. Take a look at what's happening with registering a new job. ABI is packed and input and bytecode are verified before accepting the job. I think it makes sense to do something similar here in order to avoid storing faulty contracts and wasting space.

x/evm/keeper/msg_server_user_smart_contract.go
```diff
@@ -0,0 +1,57 @@
+package keeper
+
+import (
+	"context"
+
+	"github.com/palomachain/paloma/x/evm/types"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+func (k Keeper) UploadUserSmartContract(
```
byte-bandit, line 10: FYI, since this is the `msgServer` implementation, you can use `msgServer` instead of `Keeper` as a receiver. Take a look at other modules implementation for this. Even though it's the same struct ultimately, it's a bit more explicit about intention.

proto/palomachain/paloma/evm/user_smart_contract.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+package palomachain.paloma.evm;
+
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/palomachain/paloma/x/evm/types";
+
+enum DeploymentStatus {
+  DEPLOYMENT_STATUS_UNSPECIFIED = 0;
+  DEPLOYMENT_STATUS_ONGOING = 1;
```
byte-bandit, line 10: We might want to keep it more in line with the existing lingo on smart contract deployments for this state. Take a look at their status names, I think it's like `PENDING` or `IN_FLIGHT`.

x/evm/types/keys.go
```diff
@@ -15,8 +17,19 @@ const (
 
 	// MemStoreKey defines the in-memory store key
 	MemStoreKey = "mem_evm"
+
+	UserSmartContractStoreKeyPrefix = "user-smart-contract-"
+	UserSmartContractKeyPrefix      = "contract-"
```
byte-bandit, line 22: I don't think we're using trailing hyphens anywhere else for prefixes. What's your reasoning behind that?

proto/palomachain/paloma/evm/user_smart_contract.proto
maharifu: I was assuming we only wanted to keep track of the latest deployment for each chain. I'll change it to keep track of all deployments. Including the block height should be enough for that.

> Also, what happens if a user removes the smart contract again? All deployment history will be lost as well.
Yes. It could be confusing to keep track of deployments on contracts we no longer have on chain.

x/evm/types/keys.go
maharifu: I was building the keys manually with the prefix instead of using a proper prefix store. I'll look into changing it.

x/evm/keeper/grpc_user_smart_contracts.go
maharifu: In this specific case, though, it returns a list using `keeperutil.IterAll`, which will return an empty slice instead of `ErrNotFound`, so it should be fine here.

proto/palomachain/paloma/evm/turnstone.proto
```diff
@@ -38,6 +38,17 @@ message UploadSmartContract {
   uint32 retries = 5;
 }
 
+message UploadUserSmartContract {
+  bytes bytecode = 1;
+  string abi = 2;
+  bytes constructorInput = 3;
+
+  string val_address = 4 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
```
byte-bandit, line 46: `author` might have been a more intuitive name here.

proto/palomachain/paloma/evm/tx.proto
```diff
@@ -39,3 +35,42 @@ message MsgRemoveSmartContractDeploymentRequest {
...
+  string abi_json = 3;
+  string bytecode = 4;
+  string constructor_input = 5;
+}
+
+message MsgUploadUserSmartContractResponse {
+  uint64 id = 1;
+}
+
+message MsgRemoveUserSmartContractRequest {
```
byte-bandit, line 55: The way I understand the code, a user is unable to interact (deploy/remove) with contracts that were not created by them, as they won't have access to their store. Just leaving this here for you to confirm.

byte-bandit (approved):
Looking good!

proto/palomachain/paloma/evm/tx.proto
maharifu: Yes, that's the idea. User only interact with their own contracts. :+1:

proto/palomachain/paloma/evm/turnstone.proto
maharifu: I started with `owner` but didn't like. Changed it to `author` now. Thanks :+1:

proto/palomachain/paloma/evm/set_smart_contract_deployer_proposal.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package palomachain.paloma.evm;
+
+option go_package = "github.com/palomachain/paloma/x/evm/types";
+
+message SetSmartContractDeployerProposal {
+  string title = 1;
+  string summary = 2;
+  string chainReferenceID = 3;
+  string deployerAddress = 4;
```
byte-bandit, line 10: What about turning this into a batch proposal? That way, we can set up multiple chains with one vote.

proto/palomachain/paloma/evm/user_smart_contract.proto
```diff
@@ -0,0 +1,35 @@
...
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/palomachain/paloma/x/evm/types";
+
+enum DeploymentStatus {
+  PENDING = 0;
+  IN_FLIGHT = 1;
+  ACTIVE = 2;
+  ERROR = 3;
+}
```
byte-bandit, line 13: I guess this one can be nested if not used by other message types.

x/evm/client/cli/tx_proposal.go
```diff
@@ -370,3 +371,77 @@ func CmdEvmProposalSetFeeManagerAddress() *cobra.Command {
 
 	return cmd
 }
+
+func CmdEvmProposalSetSmartContractDeployer() *cobra.Command {
```
byte-bandit, line 375: I think we should use away from legacy proposals like this. Using the established proposal flow using draft files is easier to use, and we don't have to write a new CLI endpoint for every proposal.

x/evm/keeper/attest_upload_user_smart_contract.go
```diff
@@ -0,0 +1,168 @@
...
+		if err != nil {
+			a.logger.WithError(err).Error("Failed to set UserSmartContract error")
+		}
+
+		return
+	}
+
+	a.action.Retries++
+
+	a.logger.Info("Retrying failed UploadSmartContract message",
```
byte-bandit, line 151:
```suggestion
	a.logger.Info("Retrying failed UploadUserSmartContract message",
```

proto/palomachain/paloma/evm/user_smart_contract.proto
byte-bandit: Or in case this is shared with the compass deployment flow, disregard.

x/evm/types/turnstone_abi.go
```diff
@@ -220,6 +220,85 @@ func (_m *Message_SubmitLogicCall) keccak256(
...
+		// If we have no fees set in the message, we use the same defaults as
+		// pigeon
+		fees = &Fees{
+			RelayerFee:   100_000,
+			CommunityFee: 100_000,
+			SecurityFee:  100_000,
+		}
+	} else {
+		fees = m.Fees
+	}
```
byte-bandit, line 268: Maybe extract the stuff with the fees, as the same is happening with SLC?

byte-bandit (commented):
Good stuff and great call on the fee abstraction. Some remarks, mostly about the proposal type.

PR #1203, opened 2024-06-28
chore: VerifyAgainstTX implementation

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1782

# Background

Each message type attested by Paloma needs to implement `VerifyAgainstTX` to ensure the reported evidence actually matches the expected evidence.

For each message type - `UploadSmartContract`, `UpdateValset` and `SubmitLogicCall` we implement the `VerifyAgainstTX` method. It reconstructs the transaction data from the original message in the same way as Pigeon does and compares the resulting bytecode against what's reported in the evidence.

# Testing completed

# Breaking changes


byte-bandit (approved):
Lovely stuff!

PR #1204, opened 2024-07-01
chore: print formatted queue messages on remove

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1783

# Background

At the moment, investigating messages is incredibly tedious due to the high volatility in storage. We will need a better way to investigate messages after they've been removed from the consensus queue.
This PR logs all messages as they're removed from the queue, so we can see what's happening through the logs.

- Change consensus keeper `cdc` from `BinaryCodec` to `Codec`. The `Codec` interface embeds `BinaryCodec` as well, so we don't lose functionality. However, `Codec` also adds `JSONCodec`.
- Create `MessageWithSignatures` from `QueuedSignedMessageI`.
- Marshal messages with `cdc.MarshalJSON` as we remove them from the queues. We get the output formatted the same as in the CLI commands.

# Testing completed

# Breaking changes


x/consensus/keeper/consensus/consensus.go
```diff
@@ -342,6 +343,24 @@ func (c Queue) Remove(ctx context.Context, msgID uint64) error {
 		types.ItemRemovedEventID.With(fmt.Sprintf("%d", msgID)),
 		types.ItemRemovedChainReferenceID.With(c.qo.ChainReferenceID),
 	)
+
+	logger := liblog.FromSDKLogger(sdkCtx.Logger()).WithFields("msg_id", msgID)
```
byte-bandit, line 347:
```suggestion
	logger := liblog.FromSDKLogger(sdkCtx.Logger()).WithFields("msg-id", msgID)
```

Logging fields are always [kebab-case](https://developer.mozilla.org/en-US/docs/Glossary/Kebab_case)

byte-bandit (commented):
LGTM, one small change.

byte-bandit (approved):

PR #1205, opened 2024-07-01
feat: Prevent parallel delivery of same sender msgs

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1760

# Background

This change prevents multiple messages from the same
    sender to be relayed in parallel, even by different
    relayers. This is a preliminary security step for
    pigeon feed to prevent exploitable race conditions.
    
There's also some house keeping and tidying around
    the filtering logic.

# Testing completed

# Breaking changes

### DIFF
--- a/x/consensus/keeper/filters/is_oldest_per_sender_filter.go
+++ b/x/consensus/keeper/filters/is_oldest_per_sender_filter.go
@@ -1,0 +1,27 @@
+package filters
+
+import evmtypes "github.com/palomachain/paloma/x/evm/types"
+
+// Biased: Filter expects messages to be ordered by message ID (default)
+// Filter requires injection of a look-up table to keep track of senders
+// Filter out multiple messages from same sender
+// This is done to avoid possible relays of more messages than
+// user is able to pay for.
+func IsOldestMsgPerSender(lut map[string]struct{}, msg *evmtypes.Message) bool {
+	slc := msg.GetSubmitLogicCall()
+	if slc == nil {
+		return true
+	}
+	if len(slc.GetSenderAddress()) < 1 {
+		return true
+	}
+	sender := string(slc.GetSenderAddress())
+	if _, fnd := lut[sender]; fnd {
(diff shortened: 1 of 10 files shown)


x/consensus/keeper/filters/is_oldest_per_sender_filter.go
```diff
@@ -0,0 +1,27 @@
+package filters
+
+import evmtypes "github.com/palomachain/paloma/x/evm/types"
+
+// Biased: Filter expects messages to be ordered by message ID (default)
+// Filter requires injection of a look-up table to keep track of senders
+// Filter out multiple messages from same sender
+// This is done to avoid possible relays of more messages than
+// user is able to pay for.
+func IsOldestMsgPerSender(lut map[string]struct{}, msg *evmtypes.Message) bool {
```
maharifu, line 10: Since we only care about SLC messages here, should we rename this to something more specific? Perhaps `IsOldestSLCPerSender` instead?

maharifu (commented):
Looks great, just a small nitpick.

x/consensus/keeper/filters/is_oldest_per_sender_filter.go
byte-bandit: Hmm, at first I agreed. Then I realized `IsOldestSLCPerSender` is also kind of misleading, as this gets called for all messages, and will return for all other message types that might not be an SLC or have a concept of `sender`. 

So it would be called something like `IsOldestSLCOrNonSLCMessage`, which is starting to get silly. So I suppose while not 100% accurate, I'd leave it as is unless you feel strongly otherwise.

maharifu: Yeah, that makes sense. Let's keep it simple. :+1:

maharifu (approved):
LGTM

taariq:
@byte-bandit signed commit still an issue on this one. curious.

PR #1206, opened 2024-07-03
chore: rename gravity module

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1801

# Background

As a Paloma user, I want the "gravity2" module to have a name that better represents it's functionality (it was heavily modified from the original gravity source) and I want the (cli) documentation to make clear that it is multichain instead of just for ETH.

> [!CAUTION]
> This change is breaking and will remove any data stored within the Gravity store.

# Testing completed

# Breaking changes


maharifu (approved):
Nice one!

It's a huge PR, but it looks fine. The renamed commands look OK as well. :+1:

taariq:
Wow. That was big and fast! Congrats @byte-bandit

PR #1212, opened 2024-07-05
chore: change bridge-tax to be per token

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1815

# Background

We want bridge taxes to be per token, instead of global.

- Change bridge tax setter and getter to include token
- Change bridge tax proposal to include token
- Change query RPC to return all bridge taxes, same as bridge transfer limits
- Update genesis
- Removed `excluded_token` setting as it doesn't make sense now

# Testing completed

# Breaking changes


byte-bandit (changes requested):
Excellent work.

Blocking this until we're clear on what happens to live data. We have tax data in the keeper already: https://explorer.kjnodes.com/paloma/gov/57

So that means we'll need an upgrade path to ensure the upgrade from `1.15.1` to `1.15.2` won't crash due to type differences.

My suggestion would be to simply clear the store of tax data.

maharifu:
> Excellent work.
> 
> Blocking this until we're clear on what happens to live data. We have tax data in the keeper already: https://explorer.kjnodes.com/paloma/gov/57
> 
> So that means we'll need an upgrade path to ensure the upgrade from `1.15.1` to `1.15.2` won't crash due to type differences.
> 
> My suggestion would be to simply clear the store of tax data.

Won't the store be cleared with the rename of gravity to skyway in the 1.15.2 upgrade?

I tested the upgrade in local testnet and it worked fine afterward. I'll re-do the test to make sure I have bridge tax data before the upgrade.

byte-bandit (approved):
Of course the store will be cleared anyway. LGTM.

maharifu:
> Of course the store will be cleared anyway. LGTM.

Thanks for checking.

I was just testing locally. The store works fine, but we have an issue with the proposal. Since the proposal changed, we can't decode the previous one anymore.
```
# palomad q gov proposals
Error: rpc error: code = Unknown desc = runtime error: invalid memory address or nil pointer dereference: panic
```

I'll look into it.

PR #1213, opened 2024-07-05
fix: add gravity proposals so they can be decoded

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1832

# Background

We already have gov proposals from the gravity module on chain, so we need to register them in order for them to be decoded.

We need to add them in the `gravity` proto directory, so the proto file names match the proto package, otherwise we get warnings on the command line.

The resulting gravity types reside within the `skyway/types` dir.

# Testing completed

# Breaking changes


byte-bandit (approved):
Very unfortunate, but good thing we don't need a separate module.

PR #1214, opened 2024-07-05
feat: remove retry limit from erc20 transfers

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1050
- https://github.com/VolumeFi/paloma/issues/1802

# Background

This change will make sure that ERC20 transfers are being retried continuously instead of stopping after 2 retries to try and attempt to prevent a deadlock. There might still be a need for a manual unlock down the road.

Additionally, we now make sure to stop processing skyway messages while we have an ongoing contract deployment, to avoid a desync of the nonces on the old and new compass.

# Testing completed

# Breaking changes

### DIFF
--- a/x/evm/keeper/attest_submit_logic_call.go
+++ b/x/evm/keeper/attest_submit_logic_call.go
@@ -100,10 +100,41 @@ func (a *submitLogicCallAttester) attemptRetry(ctx sdk.Context, proof *types.Sma
 		types.SmartContractExecutionMessageType.With(fmt.Sprintf("%T", a.action)),
 	)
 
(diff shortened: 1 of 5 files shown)


byte-bandit:
I ultimately decided against the watch dog approach, since we're lacking access to a lot of dynamic information at the `EndBlocker`, that is naturally available during the attestation process. So instead, we just keep brute forcing the message.

maharifu (approved):
Nice one! LGTM :+1:

taariq:
@byte-bandit needs a signed commit to merge.

PR #1218, opened 2024-07-10
fix: add valsetID to turnstone public access data

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1857

# Background

Get the valset ID from the message public access data, set by pigeon. Paloma uses this public access data when verifying the TX and get the same valset as pigeon.

- Add valsetID as new field, instead of packing it into the PublicAccessData data. This makes it more explicit and less error prone.
- When attesting SLC and UpdateValset messages we now get the same valset as pigeon (which it gets from compass), instead of the latest version store in paloma.
- Update the SetPublicAccessData so pigeon can set the valset ID.

# Testing completed

# Breaking changes


byte-bandit (approved):
lgtm

PR #1223, opened 2024-07-16
fix: message attestation with wrong sign data

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1875

# Background

UpdateValset and SubmitLogicCall messages would fail to attest if:
* Some pigeons sign the message in the queue and a pigeon picks it up
* While this pigeon is relaying the message, some more pigeons sign it
* When we try to attest it, the SignData in paloma doesn't match the SignData of the original message, since more pigeons signed it since!

To fix it, we iterate through the message SignData and remove the last signature until we get a match, or there are no more signatures.

# Testing completed

# Breaking changes

### DIFF
--- a/x/evm/types/eth_txable.go
+++ b/x/evm/types/eth_txable.go
@@ -79,36 +79,40 @@ func (m *SubmitLogicCall) VerifyAgainstTX(
 		return err
 	}
 
-	args := []any{
-		BuildCompassConsensus(valset, msg.GetSignData()),
-		CompassLogicCallArgs{
-			LogicContractAddress: common.HexToAddress(m.GetHexContractAddress()),
-			Payload:              m.GetPayload(),
-		},
-		new(big.Int).SetInt64(int64(msg.GetId())),
-		new(big.Int).SetInt64(m.GetDeadline()),
-	}
-
 	contractABI, err := abi.JSON(strings.NewReader(compass.GetAbiJSON()))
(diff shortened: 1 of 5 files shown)


x/evm/types/eth_txable.go
```diff
@@ -79,36 +79,40 @@ func (m *SubmitLogicCall) VerifyAgainstTX(
...
 
-	input, err := contractABI.Pack("submit_logic_call", args...)
-	if err != nil {
-		logger.WithError(err).Warn("SubmitLogicCall VerifyAgainstTX failed to pack ABI")
-		return err
-	}
+	// Since some validators might have added their signature to the message
+	// after a pigeon start relaying it, we iteratively remove the end signature
+	// until we get a match, or there are no more signatures.
+	for i := len(msg.GetSignData()); i > 0; i-- {
```
taariq, line 91: Does this slow down relaying itself? I'm thinking no because it's really letting relay continue, but waiting for all signatures received to repeat?

maharifu: I don't think it does.

It doesn't wait for validators to sign again, it just iterates through them. In the worst case we'll have to run this as many times as the number of validators we have, so I think it should be quick enough. The most common case will be to iterate just a couple of times.

byte-bandit (approved):

PR #1224, opened 2024-07-18
feat: lightnode sale

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1866

# Background

This PR prepares Paloma for the light node client registration. It adds:
- Paloma light node feegranter governance vote - as the light nodes will require feegrants, this vote determines who pays for the fees.
- Light node client registration transaction - to register a sold light node and start vesting
- Manual light node client license transaction - allows to create light node licenses ad-hoc, with the creator as the source of funds
- Queries for light node licenses and feegranter

NOTE: this requires an upgrade, as the paloma module did not yet have a store. It is currently set to the 1.15.4 release, but needs be changed before merging if this gets pushed to a later release.

# Testing completed

# Breaking changes


x/paloma/keeper/keeper.go
```diff
@@ -161,3 +175,201 @@ func (k Keeper) CheckChainVersion(ctx context.Context) {
...
+			if errors.Is(err, keeperutil.ErrNotFound) {
+				return types.ErrNoFeegranter
+			}
+
+			return err
+		}
+
+		// TODO - we might want to set spend and expiration limits
+		allowance := &feegrantmodule.BasicAllowance{
+			SpendLimit: nil, // Unlimited spend
```
byte-bandit, line 301: I lack oversight of the specs for this feature, so maybe I'm mistaken here. But shouldn't we avoid a situation where we give infinite fee grant to a random client address? 

What is this grant used for, and why don't we limit it?

byte-bandit (commented):
Pretty straight forward. One discussion point.

x/paloma/keeper/keeper.go
maharifu: The client account will be created at the time of registration, so it needs a fee grant or it won't have enough fees to even register. Then it starts vesting, but since it's over a two year period, it may not have enough funds to delegate for awhile.

The requirement is to not worry about abuse for now, but I think we could still set some reasonable limits (on both time and amount).

We could even limit the fee grant to only the registration message, so they could register but couldn't abuse it. They would have to wait a bit to delegate, though.

@taariq Should we set some limits here?

byte-bandit: It doesn't have to be a super low limit either. But it shouldn't be unlimited.

taariq: @maharifu why don't we put a limit of 1 GRAIN for fees. That should be enough to get the accounts to start. Right?

maharifu: @taariq Yes, I think that should be more than enough. I'll set it.

taariq (approved):
LGTM. Node Sales are hot.

taariq (approved):
Good sanity check. LFG.

byte-bandit (approved):
Perfect :)

PR #1226, opened 2024-07-19
Enable performance based relayer selection

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1220
- https://github.com/VolumeFi/paloma/issues/422

# Background

This change introduces fundamental changes to the relayer selection process. Live values are now gathered from the metrics module, and individual relayer fees are taken into account.

Going forward, it's paramount for relayers to make sure they have their fees set and configured for each chain, otherwise they will **not be eligible for relaying anymore**. 

The majority of the changes are in `msg_assigner.go`, that's where most of your attention will be needed. It works by collecting metric and fee data once per block for all validators and ranking them. Then each message during that block will be filtered depending on its requirements, and the top ranking eligible validator will be picked (and removed from the cache).

They become eligible again next block. There are still some issues with this algorithm though. At the moment, the logic will prioritize the few top ranking validators only, with little way for the mid tier to rank up. This will need to be revisited in the future for sure.

# Testing completed

# Breaking changes


x/treasury/keeper/grpc_query.go
```diff
@@ -28,3 +29,25 @@ func (k Keeper) RelayerFee(ctx context.Context, req *types.QueryRelayerFeeReques
...
 }
+
+func (k Keeper) RelayerFees(ctx context.Context, req *types.QueryRelayerFeesRequest) (*types.QueryRelayerFeesResponse, error) {
+	i, err := k.GetRelayerFeesByChainReferenceID(ctx, req.ChainReferenceId)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get relayer fees: %w", err)
+	}
+
+	response := make([]types.RelayerFeeSetting, 0, len(i))
+	for _, key := range maps.Keys(i) {
```
maharifu, line 40: It might not be relevant here, but I think this is just sidestepping the linter.
From https://pkg.go.dev/golang.org/x/exp/maps#Keys
```
The keys will be in an indeterminate order.
```
We should sort them if we really must.

x/evm/keeper/msg_assigner.go
```diff
@@ -3,266 +3,324 @@ package keeper
...
-	if len(assignableValidators) == 0 {
-		return "", errors.New("no assignable validators for message")
+	for valAddr, info := range validatorsInfos {
+		scores := map[string]math.LegacyDec{
+			"fee":           scoreValue(*pdw.maxFee, *pdw.minFee, info.Fee, true),
+			"uptime":        scoreValue(*pdw.maxUptime, *pdw.minUptime, info.Uptime, false),
+			"successRate":   scoreValue(*pdw.maxSuccessRate, *pdw.minSuccessRate, info.SuccessRate, false),
+			"executionTime": scoreValue(*pdw.maxExecutionTime, *pdw.minExecutionTime, info.ExecutionTime, true),
+			"featureSet":    scoreValue(*pdw.maxFeatureSet, *pdw.minFeatureSet, info.FeatureSet, false),
+		}
```
maharifu, line 257: Might just be personal preference, but why use a map instead of individual variables?

maharifu (commented):
Looks really good! Left a few comments but they're all just minor issues and nitpicking.

x/evm/keeper/msg_assigner.go
byte-bandit: This was from the original implementation, I can only assume the engineer back then was looking for something like dynamic flexibility with these metrics.

I just didn't touch it, but I agree this could be moved to a struct.

x/treasury/keeper/grpc_query.go
byte-bandit: It's not even sidestepping it unfortunately. My first attempt was to sort the keys, but for that we also have to iterate the map, even though it won't be consensus breaking. But CodeQL has no way of disabling warnings with annotations (yet). There's some CI extension that implements this, but I don't have the time to play around with it.

x/evm/keeper/msg_assigner.go
```diff
@@ -3,266 +3,324 @@ package keeper
...
+			// Is MEV support required?
+			if req == nil || !req.EnforceMEVRelay {
+				return true
+			}
+
+			for _, t := range v.Traits {
+				if req.EnforceMEVRelay && t == valsettypes.PIGEON_TRAIT_MEV {
+					return true
+				}
+			}
```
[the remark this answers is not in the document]
byte-bandit: Good point.

x/evm/keeper/msg_assigner.go
```diff
@@ -3,266 +3,324 @@ package keeper
...
 
-	return pickValidator(ctx, validatorsInfos, relayWeights), nil
+			// Is MEV support required?
+			if req == nil || !req.EnforceMEVRelay {
+				return true
+			}
+
+			for _, t := range v.Traits {
+				if req.EnforceMEVRelay && t == valsettypes.PIGEON_TRAIT_MEV {
```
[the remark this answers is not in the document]
byte-bandit: You're right. I think I kept this as explicit as possible to be clear about what's going on. I'd probably leave it like this unless you feel strongly otherwise.

maharifu (approved):
:+1:

PR #1229, opened 2024-07-24
feat: lightnode licenses

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1865

# Background

- Add a new governance vote to set the node sale funder account
- Process node sale events from pigeons
  - Process them in the skyway module to take advantage of the whole attestation infrastructure we already have there
  - The node sale licenses are still created in the paloma module

# Testing completed

# Breaking changes


proto/palomachain/paloma/skyway/msgs.proto
```diff
@@ -205,4 +218,26 @@ message MsgUpdateParams {
...
+// MsgLightNodeSaleClaim is sent by pigeon whenever it gets a sale notification
+// from compass
+message MsgLightNodeSaleClaim {
+  option (cosmos.msg.v1.signer) = "metadata";
+
+  palomachain.paloma.valset.MsgMetadata metadata = 1
+      [ (gogoproto.nullable) = false ];
+
+  uint64 event_nonce = 2;
+  uint64 eth_block_height = 3;
```
byte-bandit, line 233: What about

```suggestion
  uint64 remote_block_height = 3;
```

x/skyway/keeper/cosmos-originated.go
```diff
@@ -61,6 +61,22 @@ func (k Keeper) GetChainsWithTokens(ctx context.Context) []string {
 	return r
 }
 
+func (k Keeper) GetActiveChains(ctx context.Context) []string {
```
byte-bandit, line 64: Can we move this to the EVM keeper? Why are we creating a getter like this for business logic from a different module?

byte-bandit: Oh god please tell me it's not a dependency issue ...

byte-bandit (commented):
Should this live in genesis?

maharifu:
> Should this live in genesis?

It definitely should! I'll add it.

proto/palomachain/paloma/skyway/msgs.proto
maharifu: Yes, I wrote this before your PR renaming the fields - https://github.com/palomachain/paloma/pull/1228

Should I merge to that branch instead? There may be some conflicts between the two.

x/skyway/keeper/cosmos-originated.go
maharifu: It's (mostly) a visibility issue. This gets called outside the keeper, and I feel we shouldn't be exposing the `EVMKeeper` within `skyway.Keeper` just because of it.
I added a wrapper and moved the logic to the evm Keeper. Let me know what you think.

proto/palomachain/paloma/skyway/msgs.proto
maharifu: Oh, that's right, the interface is expecting this field - https://github.com/palomachain/paloma/blob/master/x/skyway/types/msgs.go#L117
So we either keep it for now, or merge your PR as well, or rename it after the PR is merged.

x/skyway/keeper/cosmos-originated.go
byte-bandit: If I didn't miss another usage, this gets called during the endblocker? 

Why not just do this:

```
chains := k.EvmKeeper.GetActiveChains(ctx)
```

That way we don't have to shallow copy the same functionality on a different keeper. I think exposing the EVM keeper here instead for usage is okay.

proto/palomachain/paloma/skyway/msgs.proto
byte-bandit: Okay, I suggest we move this into the `dev16` branch as well then. It will block rolling this out before PF is ready, but it looks like there's some overlap anyway.

maharifu: I rebased to the `dev16` branch and fixed the conflicts. However, there are still some eth mentions in the codebase, most importantly the [EthereumClaim interface](https://github.com/palomachain/paloma/blob/master/x/skyway/types/msgs.go#L107). I could refactor it as well, but it would be a big change for both paloma and pigeon. I'm not sure if we should do it now under the current time constraints, but it should be addressed in its own ticket anyway.

byte-bandit: There's nothing we gain here apart from consistency. I value consistency pretty highly, but let's circle back to this one and focus on getting the whole thing out the door first.

byte-bandit (approved):
LGTM

PR #1235, opened 2024-07-30
fix: sort validators with same score by address

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1929

# Background

We need to ensure validators are ordered even if they have the same score, otherwise the result might not be deterministic.

# Testing completed

# Breaking changes


byte-bandit (approved):
Nice one. There is a similar usage in the treasury module that should be non-blocking since it's for queries only, but I do utilise it in the PF change. I will make this change there as part of my work.

PR #1238, opened 2024-07-31
feat: add support for gas estimation on CQ messages

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/344
- https://github.com/VolumeFi/paloma/issues/1761

# Background

> [!IMPORTANT]  
> This will only merged back to master after Pigeon changes and successful smoke tests on PTN

This change adds the infrastructure for Paloma to support gas estimation. Most notably, it 

- changes the lifecycle of the consensus queue message
- adds endpoints for Pigeons to query for and report gas estimation
- uses an endblocker to process any estimated messages

## Lifecycle changes

The current message lifecycle is very bare bone. Pigeons ask for messages to be signed, and ultimately relay it. There is little filtering or state processing done. A message will even be handed out for relaying long before it has reached consensus on signatures (Pigeon checks that, not Paloma).

This change adds the concept of gas estimation to all consensus queue messages. It's not super pretty, since really only EVM messages will need the data. However, the CQ module is the main interface for managing the queue and message lifecycles. It's already grossly interwoven with the EVM module, and the whole idea of abstraction has not really paid off. We hope to address this in the future.

Paloma will hand out messages for gas estimation once they have reached consensus on signatures. Pigeons will not relay a message that requires gas estimation and has not yet received one. Even though a message might have been fully signed, it won't be relayed until gas estimation is completed. 

For this, Pigeons will query the message, build the payload and estimate the gas using their configured RPC provider. They will send their estimate to Paloma, and Paloma will start tracking every estimate. We expect estimates to be different across RPC providers, and have settled to use the median value of all collected estimates, with a generous `1.2` multiplier for safety margins.

Paloma will continuously monitor messages and check whether they have received enough estimates to calculate the estimate. IF the message happens to be an SLC message, Paloma will then proceed to automatically calculated the fees based on the relayer fee settings for this message.

Once a message has been estimated, all signatures are removed from the message and the signing process starts again, this time using the embedded information (fees & estimates). After this second round of signatures, the messages will finally be sent.

# Testing completed

# Breaking changes


x/consensus/keeper/estimate.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil
+	}
+
+	fees, err := k.calculateFeesForEstimate(ctx, sdk.ValAddress(m.GetAssignee()), m.GetChainReferenceID(), estimate)
+	if err != nil {
+		return fmt.Errorf("failed to calculate fees for estimate: %w", err)
+	}
+	action.SubmitLogicCall.Fees = fees
+	_, err = q.Put(ctx, msg, &consensus.PutOptions{
+		MsgIDToReplace: msg.GetId(),
```
maharifu, line 122: Do we need to use the same ID here? Could we drop this message and create a new one instead? Perhaps keeping the original ID in a different field.

My worry is that by using the same ID we could get some slow node to try to add it's signature or estimate to a pre-estimated message. Is this even a possibility?

x/consensus/keeper/estimate.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get fee settings: %w", err)
+	}
+
+	// Relayer fees are stored as multiplicator of total gas estimate value
+	fees.RelayerFee = multiplicators.RelayerFee.
+		MulInt(math.NewIntFromUint64(estimate)).
+		Ceil().
+		BigInt().
+		Uint64()
```
maharifu, line 145: Do we need all these int type conversions, or could we do it directly in uint64? The final result has to fit in uint64, so what are we gaining by multiplying with big types?

x/consensus/keeper/msg_server.go
```diff
@@ -39,3 +41,34 @@ func (k msgServer) AddMessagesSignatures(goCtx context.Context, msg *types.MsgAd
...
+		return nil, err
+	}
+
+	if msg.Estimates == nil {
+		return nil, nil
+	}
+
+	for _, estimate := range msg.Estimates {
+		if estimate.GetValue() < 1 {
+			return nil, fmt.Errorf("ivnalid gas estimate")
```
maharifu, line 61:
```suggestion
			return nil, fmt.Errorf("invalid gas estimate")
```

util/libcons/consensus.go
```diff
@@ -178,3 +191,57 @@ func (c ConsensusChecker) VerifyEvidence(ctx sdk.Context, evidences []*consensus
...
+			continue
+		}
+
+		logger.WithFields("value", k, "voters", v).Debug("Including value in the calculation.")
+		validEstimates = append(validEstimates, k)
+	}
+
+	// Retrieve the median value of the gas estimates and
+	// multiply value by 1.2 to allow for some security margin
+	winner := palomath.Median(validEstimates)
```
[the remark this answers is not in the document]
byte-bandit: Good idea. Done.

util/liblog/liblog.go
```diff
@@ -61,6 +62,11 @@ func (l *lgwr) WithComponent(c string) Logr {
 	return l
 }
 
+func (l *lgwr) WithChain(c string) Logr {
+	l.l = l.l.With("chain-reference-id", c)
```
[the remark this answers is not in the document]
byte-bandit: Yes, you're correct. When I wrote this lib, I  was trying to add some sanity to the logging chaos across the project, mostly to establish fixed terms to query and build filters for. I find log correlation still massively lacking. Usually, I would often end up storing log fields in a short lived table lookup and store the hash on the context, so function calls further down the chain can easily retrieve a fully contextualised logger.

But I didn't have a lot of time when I wrote this, and just sort of built an abstraction layer to enforce some standards in logging. This has created 2 problems.

1. I cannot actually retrieve a contextualised logger from the context and have to recreate it every time i want to use it. Bad.
2.  The underlying implementation doesn't really work with a `map[string]any`, so every time I want to update a field (i.e. in a loop like you found), it appends a field instead of creating a new one.

In the future, I  want to overhaul this to allow proper contextualised logging. Until then, we need to be careful inside loops. I'll take a look at the changes.

x/consensus/keeper/concensus_keeper.go
```diff
@@ -457,6 +498,40 @@ func (k Keeper) AddMessageSignature(
...
 }
 
+func (k Keeper) AddMessageGasEstimates(
+	ctx context.Context,
+	valAddr sdk.ValAddress,
+	msgs []*types.MsgAddMessageGasEstimates_GasEstimate,
+) error {
+	for _, msg := range msgs {
+		sdkCtx := sdk.UnwrapSDKContext(ctx)
+		cq := whoops.Must(
```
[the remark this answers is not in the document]
byte-bandit: This is pretty much a copy/paste from `AddMessageEvidence`. I need to make some sacrifices to keep moving, so I prefer to have them all work the same for now.

x/consensus/keeper/estimate.go
byte-bandit: This is fully possible, and the current architecture sort of anticipates this. I think recreating the message will be more painful though. In case a new estimate comes in, it will just be disregarded. Same for a new signature against the old state. But this is already the case today, so it's not like we lose something.

x/consensus/keeper/estimate.go
byte-bandit: Since `float64` is not deterministic, the `multiplicators` are stored as `LegacyDec`.  The casts are needed to work against the type interface unfortunately.

maharifu (approved):
Nice work! LGTM

PR #1245, opened 2024-08-06
feat: add fee manager to consensus

byte-bandit:
# Background

The fee manager smart contract will be independent from compass and the bridge. It will need to be deployed manually before compass, and ownership will have to be passed to compass after its deployment. 

> [!CAUTION]
> There is no support yet for automatic compass upgrades, as they will require a handover of the fee manager just like ERC20 tokens. This will be added in a follow up MR.

# Testing completed

# Breaking changes


maharifu (approved):
LGTM :+1:

PR #1246, opened 2024-08-06
chore: disable token relink before compass 2.0

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1891

# Background

Even if we don't want to move test tokens, since we have them set, paloma will still try to move them. This will fail since (new) pigeon won't be compatible with (old) compass, so we need to disable this functionality before compass 2.0 is deployed and re-enable it again afterwards.

# Testing completed

# Breaking changes

### DIFF
--- a/x/evm/keeper/attest_test.go
+++ b/x/evm/keeper/attest_test.go
@@ -47,15 +47,15 @@ var (
 	})
 )
 
-type record struct {
-	denom string
-	erc20 string
-	chain string
-}
+// type record struct {
+// 	denom string
+// 	erc20 string
+// 	chain string
+// }
 
-func (r record) GetDenom() string            { return r.denom }
-func (r record) GetErc20() string            { return r.erc20 }
-func (r record) GetChainReferenceId() string { return r.chain }
+// func (r record) GetDenom() string            { return r.denom }
+// func (r record) GetErc20() string            { return r.erc20 }
+// func (r record) GetChainReferenceId() string { return r.chain }
 
 func TestKeeperGinkgo(t *testing.T) {
 	RegisterFailHandler(g.Fail)
@@ -114,7 +114,7 @@ var _ = g.Describe("attest router", func() {
 	var ctx sdk.Context
 	var q *consensusmocks.Queuer
 	var v *evmmocks.ValsetKeeper
-	var gk *evmmocks.SkywayKeeper
+	// var gk *evmmocks.SkywayKeeper
 	var consensukeeper *evmmocks.ConsensusKeeper
 	var mk *evmmocks.MetrixKeeper
 	var tk *evmmocks.TreasuryKeeper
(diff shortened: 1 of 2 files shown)


byte-bandit:
@maharifu Instead of de- and re-activating this during the upgrade path, what about if we removed support for the chains and re-added them instead?

maharifu:
> @maharifu Instead of de- and re-activating this during the upgrade path, what about if we removed support for the chains and re-added them instead?

That would be cleaner, but it would also take longer, since we have to do by governance vote, right? Also, are there any other side effects from removing support for a chain?

byte-bandit:
> > @maharifu Instead of de- and re-activating this during the upgrade path, what about if we removed support for the chains and re-added them instead?
> 
> That would be cleaner, but it would also take longer, since we have to do by governance vote, right? Also, are there any other side effects from removing support for a chain?

I'm not 100% sure tbh. Messages should be drained automatically, but I'm unclear about deployed tokens and whether they're cleaned up or not. Would need some investigation.

You're right about the additional downtime though :/

maharifu:
> > > @maharifu Instead of de- and re-activating this during the upgrade path, what about if we removed support for the chains and re-added them instead?
> > 
> > 
> > That would be cleaner, but it would also take longer, since we have to do by governance vote, right? Also, are there any other side effects from removing support for a chain?
> 
> I'm not 100% sure tbh. Messages should be drained automatically, but I'm unclear about deployed tokens and whether they're cleaned up or not. Would need some investigation.
> 
> You're right about the additional downtime though :/

I'll try to test this in the private testnet and see what happens. It's not a guarantee, but should highlight obvious issues, if any. We can leave the PR open in the meanwhile.

maharifu:
@byte-bandit The ERC20 tokens are independent of the supported chains, so removing support for a chain does not remove the ERC20 tokens.

Worse still, there's currently a logic bug in the token relink function, such that if we only have ERC20 tokens for unsupported chains when trying to attest a new compass upload, we will [start to relink](https://github.com/palomachain/paloma/blob/master/x/evm/keeper/attest_upload_smart_contract.go#L110), but [not do anything](https://github.com/palomachain/paloma/blob/master/x/evm/keeper/attest_upload_smart_contract.go#L150). The deployment cache [won't have the msgID](https://github.com/palomachain/paloma/blob/master/x/evm/keeper/attest_submit_logic_call.go#L51) and the contract won't become active.

We can create a new vote to remove ERC20 tokens, which we will probably need anyway in the future, but it would need a new release and a new vote.

byte-bandit (approved):
Let's take the pain then. Can you setup a reminder ticket so we don't forget to re-enable this later on please? Also worth taking a note of the bug you found.

maharifu:
> Let's take the pain then. Can you setup a reminder ticket so we don't forget to re-enable this later on please? Also worth taking a note of the bug you found.

Sure. I already have a ticket for the bug - https://github.com/VolumeFi/paloma/issues/1953 - and created a new one to revert this PR - https://github.com/VolumeFi/paloma/issues/1956

PR #1247, opened 2024-08-07
fix: jit valset updates in SLCs

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/1042

# Background

Before issuing a jit valset update, check the queue for existing valset updates with the same ID.

# Testing completed

# Breaking changes


byte-bandit (changes requested):
Can we merge directly do `dev16`?

byte-bandit (approved):

PR #1248, opened 2024-08-07
chore: update gas estimate query response

byte-bandit:
# Background

Small change needed on the contract between Paloma/Pigeon.

# Testing completed

# Breaking changes


byte-bandit:
@maharifu FYI

PR #1250, opened 2024-08-08
include estimate in message query result

byte-bandit:
# Testing completed

# Breaking changes


maharifu (approved):
:+1:

PR #1257, opened 2024-08-12
chore: gas estimates in sign bytes

maharifu:
# Background

Include gas estimate and fee structure in turnstone messages hash.

# Testing completed

# Breaking changes


x/evm/types/turnstone_abi.go
```diff
@@ -98,19 +141,45 @@ func (_m *Message_SubmitLogicCall) keccak256(orig *Message, nonce uint64) []byte
...
+			CommunityFee: 100_000,
+			SecurityFee:  100_000,
+		}
+	} else {
+		fees = m.Fees
+	}
+
+	// Left-pad the address with zeroes
+	padding := bytes.Repeat([]byte{0}, 32-len(m.SenderAddress))
+	senderAddress := [32]byte(append(padding, m.SenderAddress...))
```
byte-bandit, line 181: Pigeon already performs this bit: https://github.com/palomachain/pigeon/blob/02c99aea1e69913a6b026a8cbeef718cf7bad4f2/chain/evm/compass.go#L257

But I guess it won't hurt.

x/evm/types/turnstone_abi.go
```diff
@@ -68,12 +72,45 @@ func (_m *Message_UpdateValset) keccak256(orig *Message, _ uint64) []byte {
...
+	estimate := gasEstimate
+	if estimate == 0 {
+		// If there's no estimate, we use the same default as pigeon
+		estimate = 300_000
+	}
 
+	bytes, err := arguments.Pack(
+		hash32,
+		common.HexToAddress(orig.AssigneeRemoteAddress),
+		new(big.Int).SetInt64(int64(gasEstimate)),
```
byte-bandit, line 106: The estimate is always stored as `uint` I believe.

byte-bandit (changes requested):
Great work, only some small remarks. There is a panic as well where there shouldn't be.

x/evm/types/turnstone_abi.go
maharifu: We need to do the same as pigeon here, so the hashes match, right?

x/evm/keeper/keeper.go
```diff
@@ -168,12 +168,40 @@ func (k *Keeper) AddMessageConsensusAttestedListener(l metrixtypes.OnConsensusMe
 	k.onMessageAttestedListeners = append(k.onMessageAttestedListeners, l)
 }
 
-func (k Keeper) PickValidatorForMessage(ctx context.Context, chainReferenceID string, requirements *xchain.JobRequirements) (string, error) {
+func (k Keeper) PickValidatorForMessage(
```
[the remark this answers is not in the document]
maharifu: I tried that, but unless we declare the type somewhere awkward, it leads to dependency cycles. I think it's best to leave it as is for now.

maharifu:
> Great work, only some small remarks. There is a panic as well where there shouldn't be.

Regarding the panics, we already have other panics in the same functions, I just continued the same style. I think we can refactor them out and return an error instead. However, I don't see these functions being used in Begin or End blockers. Is this an issue or a false positive?

x/evm/types/turnstone_abi.go
byte-bandit: Yes, yes we do ...

x/evm/keeper/keeper.go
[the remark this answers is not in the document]
byte-bandit: Yeah, we need like a `common` types package at some point, the dependencies are getting out of hand. Okay, let's leave like this for sake of speed then.

byte-bandit:
> > Great work, only some small remarks. There is a panic as well where there shouldn't be.
> 
> Regarding the panics, we already have other panics in the same functions, I just continued the same style. I think we can refactor them out and return an error instead. However, I don't see these functions being used in Begin or End blockers. Is this an issue or a false positive?

Unless I'm mistaken, the `panic` statements are in hot code paths. Did a short scan and found one in the `keccak256` for submit logic call and valset update for example, which are currently both used to calculate the bytes to sign. 

It's interesting the CodeQL plugin didn't find these yet, but we definitely want them out. In theory, failing to pack the arguments should never happen, but better be safe then risk an app hash.

maharifu:
> > > Great work, only some small remarks. There is a panic as well where there shouldn't be.
> > 
> > 
> > Regarding the panics, we already have other panics in the same functions, I just continued the same style. I think we can refactor them out and return an error instead. However, I don't see these functions being used in Begin or End blockers. Is this an issue or a false positive?
> 
> Unless I'm mistaken, the `panic` statements are in hot code paths. Did a short scan and found one in the `keccak256` for submit logic call and valset update for example, which are currently both used to calculate the bytes to sign.
> 
> It's interesting the CodeQL plugin didn't find these yet, but we definitely want them out. In theory, failing to pack the arguments should never happen, but better be safe then risk an app hash.

You are right! I'll refactor that out and return an error instead.

byte-bandit (approved):

PR #1258, opened 2024-08-12
feat: add mutable BytesToSign to skyway module

byte-bandit:
# Background

This adds mutable BytesToSign to the skyway module, they get refreshed after gas estimate is attached to the batch.

# Testing completed

# Breaking changes


x/skyway/types/batch.go
```diff
@@ -235,6 +245,7 @@ func (i InternalOutgoingTxBatch) GetChainReferenceID() string {
 
 // GetCheckpoint gets the checkpoint signature from the given outgoing tx batch
 func (i InternalOutgoingTxBatch) GetCheckpoint(turnstoneID string) ([]byte, error) {
+	// TODO Add relayer and gas estimate
```
byte-bandit, line 248:
```suggestion
```

maharifu (approved):
LGTM

PR #1275, opened 2024-08-27
Add 

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2061

# Background

- Fixes the Skyway nonce reset implementation during a compass ugprade
- Introduces a governance proposal message which may be used to override the last nonce across all validators for a given chain to recover from a desync

# Testing completed

# Breaking changes

### DIFF
--- a/x/skyway/abci.go
+++ b/x/skyway/abci.go
@@ -42,6 +42,10 @@ func EndBlocker(ctx context.Context, k keeper.Keeper, cc *libcons.ConsensusCheck
 		if err != nil {
 			logger.WithError(err).Warn("Failed to prune attestations.")
 		}
+		err = k.IterateValidatorLastEventNonces(ctx, v, func(key []byte, val uint64) bool {
+			logger.WithFields("chain-reference-id", v, "validator", key, "nonce", val).Debug("Iterating over validator last event nonces")
+			return false
+		})
 	}
 
 	err = processGasEstimates(ctx, k, cc)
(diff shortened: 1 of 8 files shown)


x/skyway/abci.go
```diff
@@ -42,6 +42,10 @@ func EndBlocker(ctx context.Context, k keeper.Keeper, cc *libcons.ConsensusCheck
 		if err != nil {
 			logger.WithError(err).Warn("Failed to prune attestations.")
 		}
+		err = k.IterateValidatorLastEventNonces(ctx, v, func(key []byte, val uint64) bool {
+			logger.WithFields("chain-reference-id", v, "validator", key, "nonce", val).Debug("Iterating over validator last event nonces")
+			return false
+		})
```
byte-bandit, line 48:
```suggestion
```

x/skyway/keeper/keeper.go
```diff
@@ -474,3 +449,30 @@ func (k Keeper) SetAllLighNodeSaleContracts(
 
 	return nil
 }
+
+func (k Keeper) overrideNonce(ctx context.Context, chainReferenceId string, nonce uint64) error {
```
byte-bandit, line 453:
```suggestion
// purposefully circumvents keeper setters and getters, as those perform some biased operations and may end up inserting 
// a different value or dropping it altogether.
func (k Keeper) overrideNonce(ctx context.Context, chainReferenceId string, nonce uint64) error {
```

PR #1277, opened 2024-08-29
Feature/compass handover

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1960
- https://github.com/VolumeFi/paloma/issues/1951
- https://github.com/VolumeFi/paloma/issues/1956
- https://github.com/VolumeFi/paloma/issues/2043

# Background

This change makes use of the atomic handover endpoint on compass, re-enables the token relink and ensures ownership of fee manager is transferred as well.

# Testing completed

# Breaking changes


taariq (approved):
Well done and alot of ground covered. Let's ship and test in prod!

PR #1292, opened 2024-09-13
chore: update validator nonces in endblocker

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2160

# Background

The way it was before, we would need a claim to be observed so validators would be updated. By doing it in the endblocker, we validators are always up-to-date.

This should prevent a situation where the last claim is 19, the last observed nonce is 18, but validators are stuck on nonce 17. They would receive a list of block events with the block of nonce 19 (latest unobserved) but would be unable to claim it, since skyway would expect them to submit a claim for nonce 18 instead.

# Testing completed

# Breaking changes

### DIFF
--- a/x/skyway/abci.go
+++ b/x/skyway/abci.go
@@ -42,6 +42,14 @@ func EndBlocker(ctx context.Context, k keeper.Keeper, cc *libcons.ConsensusCheck
 		if err != nil {
 			logger.WithError(err).Warn("Failed to prune attestations.")
 		}
+
+		// Update all validator nonces to the latest observed nonce, if suitable
+		// This makes sure validators don't get stuck waiting for events they
+		// can't get from the RPC, even though these events are already observed
+		err = k.UpdateValidatorNoncesToLatest(ctx, v)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


x/skyway/abci.go
```diff
@@ -42,6 +42,14 @@ func EndBlocker(ctx context.Context, k keeper.Keeper, cc *libcons.ConsensusCheck
...
 			logger.WithError(err).Warn("Failed to prune attestations.")
 		}
+
+		// Update all validator nonces to the latest observed nonce, if suitable
+		// This makes sure validators don't get stuck waiting for events they
+		// can't get from the RPC, even though these events are already observed
+		err = k.UpdateValidatorNoncesToLatest(ctx, v)
+		if err != nil {
+			logger.WithError(err).Warn("Failed to update validator nonces.")
+		}
```
byte-bandit, line 52: Since this isn't mission critical, let's extract this into something that runs i.e. every 50 blocks or so to keep the block speed low.

maharifu: Thank. I added that condition and tested it in the private testnet.

PR #1302, opened 2024-09-27
feat: upgrade to Cosmos SDK v0.50.10

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2092

# Background

This change upgrades to `comsos-sdk v0.50.10` and introduces support for `cosmwasm 2.0`

# Testing completed

# Breaking changes


cmd/palomad/root.go
```diff
@@ -207,3 +218,13 @@ func rootPreRunE(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+var tempDir = func() string {
```
byte-bandit, line 222: I'm aware this is some of the most horrible code you've seen in a while. It's taken straight from Cosmos SDK's reference implementation, and could likely be improved upon if I had more time. At least it works.

cmd/palomad/root.go
```diff
@@ -207,3 +218,13 @@ func rootPreRunE(cmd *cobra.Command, args []string) error {
...
+	dir, err := os.MkdirTemp("", ".paloma")
+	if err != nil {
+		panic(fmt.Sprintf("failed creating temp directory: %s", err.Error()))
+	}
+	defer os.RemoveAll(dir)
```
maharifu, line 227: Is this call supposed to be here? We also remove try to remove it on line 45. But doesn't this just remove it after returning?

maharifu (commented):
Nice work. Just left a question on the temp dir.

cmd/palomad/root.go
byte-bandit: It's a carbon copy of the the original code. Maybe it's not needed, and the calling code will recreate it? I'm not sure. I'd rather find a way to not use this hacky approach at all.

PR #1307, opened 2024-10-01
fix: append v2 to module name

maharifu:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2136

# Background

From v2 onwards, go forces us to either:
- create a new directory for every version and maintain all versions within the codebase
- append the version to the module name

We opted for the latter since we don't want to support multiple paloma versions.

# Testing completed

# Breaking changes


taariq (approved):
LFG!

PR #1309, opened 2024-10-08
feat: store lightnode activations

maharifu:
# Related Github tickets

- Closes https://github.com/VolumeFi/paloma/issues/2250
- Closes https://github.com/VolumeFi/paloma/issues/2251
- Closes https://github.com/VolumeFi/paloma/issues/2252

# Background

Keep an on-chain list of activated lightnodes. When a new node is activated, we add it to the list. To bootstrap the list with the existing lightnodes, we add a new transaction. This is instead of a governance vote, so we don't waste more time. The transaction can be executed at any time after the upgrade.
For each lightnode we keep track of: address, activated timestamp, last authentication timestamp. This way we can also know which nodes are active and performing authentications.

# Testing completed

# Breaking changes


taariq (approved):
Looks good and what we need! LFG!

PR #1318, opened 2024-12-13
feat: add token factory module

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2440

# Background

This change adds a token factory module that extends Paloma with the following features:

- New tokens may be created by anyone
- Tokens can only be minted by the admin of the token (by default, the creator of a token is the admin)
- Tokens can only be burned by the admin of the token
- All data is respected for genesis import/export
- Skyway registrations against user tokens do not require governance
- Tokens can be bridged outside of Paloma
- Creating a token costs 10 GRAIN

# Testing completed

> [!WARNING]  
> While most of the code is greatly inspired by the COSMWASM tokenfactory module and should be battle tested, I did not have time to include proper e2e tests to cover all Paloma customizations. Successful bridging still needs smoke testing.

# Breaking changes


taariq (approved):
LFG

PR #1321, opened 2024-12-30
feat: experimental wasm bindings for token factory & scheduler

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2459

# Background

This change adds experimental bindings for the token factory, as well as a limited scope of the scheduler module, allowing jobs to be queried and created from smart contracts. 

Bindings are untested and should be treated as experimental.

# Testing completed

# Breaking changes

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -750,11 +750,11 @@ func New(
 			)
 		}),
 	}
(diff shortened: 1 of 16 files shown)


taariq (approved):
Big change. Big impact. LFG and test in prod.

PR #1323, opened 2025-01-02
feat: add static binary build in Dockerfile

mamirpanah:
This PR aims to add static binary build in Dockerfile


taariq:
Thank you for the PR @mamirpanah. We'll take a look and review!

byte-bandit (changes requested):
Thanks for your contribution. Left a few comments to improve.

I'm curious, does this solution actually result in a working container that lets you run a full node?

PR #1327, opened 2025-01-14
feat: add fee skipping

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2466

# Background

This change introduces as custom `TxFeeChecker` implementation, which simply returns an empty fee for every transaction, as well as a constant priority.

Merging this change will lead to an equal transaction prioritization across Paloma, without the option to incentivize validators to include a transaction in an earlier block.

# Testing completed

# Breaking changes


taariq (approved):
LFG!

PR #1328, opened 2025-01-16
chore: enable optimistic execution

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/1162

# Background

Enforces optimistic execution across all nodes. PTN tests are looking good, but our mileage may vary depending on cross chain chatter.

# Testing completed

# Breaking changes


taariq (approved):
@byte-bandit Is this all? No other changes needed?

byte-bandit:
@taariq We're coasting along on the stock CosmosSDK implementation, which is derived from the original Sei contribution. It doesn't look like we need any custom modifications since Paloma is pretty much stock cometbft.

PR #1329, opened 2025-01-16
chore: remove legacy binding field

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2497

# Background

Removes an unused field from the scheduler module legacy bindings. Additional fields are discarded, so backwards compatibility is not broken here.

# Testing completed

# Breaking changes


taariq (approved):
ACK.

PR #1332, opened 2025-01-22
feat: add selective gas exemption whitelist

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2487

# Background

This change introduces a new governance concept: a whitelist of gas exempt addresses. The list can be updated by creating a `MsgUpdateParams` vote on the `paloma` module.

Every member address of the list will be exempt from paying gas fees on the chain.

The change also brings some QOL improvements in the integration testing package.

# Testing completed

# Breaking changes


taariq (approved):
ACK. Let's lock and load!

PR #1333, opened 2025-01-24
fix: wasm binding integrations

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2531

# Background

This change addresses a bug in the way custom `wasm` messages are parsed and routed to target modules. 

# Testing completed

# Breaking changes


taariq (approved):
ACK.

PR #1336, opened 2025-01-28
fix (scheduler): inject expected wrapping in payload

byte-bandit:
# Related Github tickets

fixes volumefi/paloma#2543

# Testing completed

# Breaking changes


webelf101:
I am curious if we need similar encoding in create_job? @byte-bandit

taariq (approved):
ACK

byte-bandit:
@webelf101 It should not be a problem, AFAICT, this wrapping is only used for actual job execution.

PR #1339, opened 2025-02-10
feat: add one-time grain restoration vote

byte-bandit:
# Related Github tickets

https://github.com/VolumeFi/paloma/issues/2570#issuecomment-2648084339- 

# Background

This change introduces a new vote which can be used to recreate the 1.3b GRAIN tokens which are lost forever on BASE. They will be funded to the Paloma Foundation wallets.

# Testing completed

# Breaking changes


taariq (approved):
ACK! Let's COO!

PR #1340, opened 2025-02-10
feat: add singular vote to update GRAIN token maps

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2573

# Background

Introduces a new vote which allows for bulk editing of the ERC20 to denom mapping in the Skyway module.

# Testing completed

# Breaking changes


taariq (approved):
ACK. Let's COO!

PR #1346, opened 2025-03-05
fix: prevent mapping multiple denoms to same ERC20

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2641

# Background

This change adds a check that prevents binding a `denom` to an ERC20 address that is already bound.

> [!important]
> This will prevent updating your ERC20 addresses against your own denoms even, but it could theoretically be implemented at a later stage.

> [!warning]
> The underlaying gravity bridge code was not written with more than one remote chain in mind. Therefore, after this change, it will not be possible to use the same Ethereum address for a denom on more than one remote chain, so deterministic addresses cannot be used anymore.

# Testing completed

# Breaking changes


taariq (approved):
ACK. Smol change. Big impact.

PR #1350, opened 2025-03-14
chore: upgrade to CosmosSDK v0.50.13

byte-bandit:
# Testing completed

# Breaking changes


taariq (approved):
@byte-bandit what about CometBFT? Do we get an upgrade there as well?

byte-bandit:
@taariq Nope, Cosmos has not moved on yet. They're also still on `go 1.21` so I won't upgrade to `1.24`. Keep the focus on stability now.

PR #1354, opened 2025-03-21
feat: add governance vote to update reference block hash

byte-bandit:
# Related Github tickets

- https://github.com/VolumeFi/paloma/issues/2726

# Background

This change adds a governance vote to override the current reference block hash. This is meant as a fail safe to recover from an incorrect hash which might have originated during a chain fork.

# Testing completed

# Breaking changes


taariq (approved):
Yes, but this means this chain can be dead for 24-hours before the vote so no trading for 24 hours like we have today.

